Issue #2, opened 2019-01-17
.vault_pass support


retlehs:
`trellis new` should allow for creating a `.vault_pass`, setting the path to the vault_pass in `ansible.cfg`, and encrypting the files

swalkinshaw:
This was mostly taken care of by #21 except for the actual encrypting part.

trellis-cli should offer more `vault` commands though.

swalkinshaw:
Closed by https://github.com/roots/trellis-cli/pull/27

PR #21, opened 2019-01-30
Generate .vault_pass file

swalkinshaw:
Also sets the `vault_password_file` in `ansible.cfg` by default.

This makes it way easier to get started with Vault but doesn't change anything by default. It's completely opt-in to actually encrypt the files.

### DIFF
--- a/cmd/new.go
+++ b/cmd/new.go
@@ -125,6 +125,16 @@ func (c *NewCommand) Run(args []string) int {
 		c.trellis.WriteVaultYaml(vault, filepath.Join("group_vars", env, "vault.yml"))
 	}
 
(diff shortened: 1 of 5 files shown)


swalkinshaw:
This mostly closes #2, not sure we should ever encrypt files by default at the start?

May be better to have a `trellis vault encrypt` to explicitly do it.

knowler:
Could we add a flag to set a custom path for this? I often store mine in Keybase and I'd find it helpful.

I imagine usage would be something like this:
```shell
$ trellis new example.com --vault-pass-path=/keybase/teams/company/sites/example.com
```

PR #24, opened 2019-02-01
Add check command for requirements

swalkinshaw:
`trellis check` will check if the required dependencies are met or not

image

image

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -1,0 +1,92 @@
+package cmd
+
+import (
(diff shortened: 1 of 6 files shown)


swalkinshaw:
Right now these requirements are hardcoded in here. I was thinking one solution is to commit a `.trellis.yml` config in Trellis itself where'd store the requirements then trellis-cli could read them. The upside is the requirements are tied to Trellis and not the CLI.

PR #28, opened 2019-02-06
DigitalOcean support

swalkinshaw:
This adds a `droplet` command with one subcommand: `create`

It will create a DigitalOcean droplet (server) via their API and automatically provision it.

A `DIGITALOCEAN_ACCESS_TOKEN` environment variable is needed to run this command. If it's not found, a prompt will export it.

This command also requires an SSH key to be added to your DO account. By default it will look in `~/.ssh/id_rsa.pub` and add it to your account if it's not already there. A different key path can be specified with the `--ssh-key` option.

trellis-cli will create a Ubuntu 18.04 x64 server. The region and size of the droplet is up to the user either via options or prompt selections.

The name of the droplet will default to the first site name in your Trellis `wordpress_sites`. This can be changed via a prompt as well.

The droplet will be created with two tags:
* `trellis`
* environment name

Once the server is created, `ansible-galaxy install` will be run to ensure the Galaxy roles exist. Then the server will be provisioned via the CLI.


artshostak:
Thanks Scott, but would definitely love for this to have some documentation, kind of scattered the pieces together but still getting an error when executing `trellis droplet create` command:

1. Installed DO's command line tool: `brew install dctl`
2. Authorized it with an API token for the project: `doctl auth init`
3. Entered the access token and validated OK
4. Now running `trellis droplet create` produces the following error:

```$ trellis droplet create
panic: runtime error: index out of range [0] with length 0
goroutine 1 [running]:
trellis-cli/cmd.(*DropletCreateCommand).Run(0xc00025f420, 0xc000112190, 0x0, 0x0, 0xc000129a80)
	/Users/scottwalkinshaw/dev/trellis-cli/cmd/droplet_create.go:75 +0xf78
github.com/mitchellh/cli.(*CLI).Run(0xc00015fcc0, 0xc000129c80, 0x16363a5, 0x9)
	/Users/scottwalkinshaw/go/pkg/mod/github.com/mitchellh/cli@v1.1.2/cli.go:262 +0x1cf
main.main()
	/Users/scottwalkinshaw/dev/trellis-cli/main.go:158 +0x13c5```

Trellis cli version: 0.9.1

**Comment by swalkinshaw** [comment]
The command requires an "environment" argument. For example:

```bash
trellis droplet create production
```

But it's a bug that the args weren't validated properly so I'll fix that 😄

swalkinshaw:
Fixed in https://github.com/roots/trellis-cli/pull/141

artshostak:
Yes king!

PR #33, opened 2019-02-14
Fix #32 - Remove special chars from random strings

### LINKED ISSUE #32: Generated secrets/password character issues
The generated passwords/keys/secrets have potential issues with special characters and not being properly quoted.

MySQL root password is definitely susceptible to this. Example: https://twitter.com/andreaDuquette/status/1095821506091892736

One issue is that the Go YAML library doesn't think it needs to quote string values unless in some conditions.

Options:
* somehow force them to be quoted
* drop some special characaters

swalkinshaw:
Special characters can cause issues in a few places:

* YAML files/templates
* Certain commands/software such as Ansible's `mysql_db`

While it may seem less secure to generate passwords without these
characters (ie: easier to brute-force), password length is still the most
important factor. Special characters are often used to ensure *human*
created passwords have more entropy (less random); however, these
strings are cryptographically randomly generated so they don't suffer
from these human issues.

So this should cause less isuess while still being secure enough for our
use cases.

Caveat: I don't really know anything about cryptography.

### DIFF
--- a/trellis/vault.go
+++ b/trellis/vault.go
@@ -118,7 +118,7 @@ func generateRandomBytes(n int) ([]byte, error) {
 }
 
 func generateRandomString(n int) string {
-	const letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%^&*()-_[]{}<>~`+=,.;:/?|"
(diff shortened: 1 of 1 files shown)


swalkinshaw:
@austinpray what do you think

Issue #7, opened 2019-01-17
Database access command


retlehs:
https://github.com/TypistTech/vagrant-trellis-sequel exists for vagrant, but we could also support sequel connections to the remote environments

TangRufus:
To do so, we need a way to _compile_ [`site_env`](https://github.com/roots/trellis/blob/faaf213ae2186c6ccf55159becd8910e20c82bf1/group_vars/all/helpers.yml#L12):
  1. decrypt vault files
  2. combine site_env
  3. handling ansible vaibles, example:
     ```
     vault_wordpress_sites:
       example.com:
         env:
            db_password: "{{ my_prefix }}_dbpassword"
     ```

Any idea how to implement them in golang?

swalkinshaw:
Your vagrant plugin doesn't do that does it? I looked through it recently and didn't see anything like that.

Evaluating `site_env` would be most impossible I think? (without just trying to replicate the logic of course). We'd need to parse that file as YAML, then have a Jinja2 parser (which isn't even enough because I think that has custom Ansible filters)... and then eval it in the context of our variables as you mentioned.

TangRufus:
> Your vagrant plugin doesn't do that does it? 

The plugin only decrypts vault files but not evaluating `site_env` nor handling ansible varibles.

It assumes users don't customize any `db_xxx` values - https://github.com/TypistTech/vagrant-trellis-sequel/blob/cc93696de180ba8e72e80039c9837f0de26bd77c/lib/vagrant_plugins/trellis_sequel/vault.rb#L26-L58

As we are going to support more than vagrant, we should expect users to use all ansible features on their `db_xxx` values.

Possible solutions:
1. SSH into remote servers + running PHP scripts (or wp cli commands) to get `db_xxx` values.
2. SSH into remote servers + parse site `.env`
3. Use ansible-playbook to generate `db_xxx` values, example: https://github.com/ItinerisLtd/enveigle

swalkinshaw:
The 3rd option seems best to me? Especially since as already proven by your enveigle plugin that functionality on its own is useful. trellis-cli could have a `dotenv`

swalkinshaw:
This has been done for a while since https://github.com/roots/trellis-cli/pull/72

PR #50, opened 2019-08-03
Add `dotenv` command

TangRufus:
Ported from https://github.com/ItinerisLtd/enveigle/tree/8f1c803b7125694516e9de15c95874cf80b99917

Related: https://github.com/roots/trellis-cli/issues/7#issuecomment-515650864

FileIO and anisble playbook (`dotenv.yml`) are not tested. Help wanted!

CircleCI is not building. Please check https://discuss.circleci.com/t/pull-requests-not-triggering-build/1213/5


TangRufus:
FileIO and anisble playbook (`dotenv.yml`) are not tested. Help wanted!

cmd/dot_env.go
```diff
@@ -0,0 +1,102 @@
...
+	if err := c.Trellis.LoadProject(); err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	var environment string
+
+	switch len(args) {
+	case 0:
+		environment = "development"
```
swalkinshaw: Oh good idea. There's probably other commands where this can be done too 👍

cmd/dot_env.go
```diff
@@ -0,0 +1,102 @@
...
+	if dotEnv.Env != nil {
+		env = dotEnv.Env
+	}
+	dotEnv.Env = append(env, "ANSIBLE_RETRY_FILES_ENABLED=false")
+
+	logCmd(dotEnv, c.UI, true)
+	runErr := dotEnv.Run()
+
+	// Remove playbook file no matter command success or not
+	removeFileErr := os.Remove(playbookPath)
```
swalkinshaw: I think we'd want this to be done via `defer` for more guarantees:

```go
defer deletePlaybook(playbookPath) # function can do error handling
```

cmd/dot_env.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// runErr gets priority because it is more important than removeFileErr
+	if runErr != nil {
+		log.Fatal(runErr)
```
swalkinshaw: The other `log.Fatal`s are probably fine to stay since they should be uncommon, but this one should use `c.UI.Error` I think

cmd/dot_env.go
TangRufus: done.

cmd/dot_env.go
TangRufus: done.
Also added `return 1` when `ansible-playbook` fails

Note the command output and the actual exit status:

```sh-session
$ trellis dotenv
// ...
failed: [192.168.50.5] (item=insecure.com) => {"changed": false, "item": {"key": "insecure.com", "value": {"local_path": "../bedrock-insecure", "site_hosts": [{"canonical": "insecure.test"}], "ssl": {"enabled": false}}}}

PLAY RECAP *********************************************************************
192.168.50.5               : ok=0    changed=0    unreachable=0    failed=1

Error running ansible-playbook: exit status 2

$ echo $?
1
```

swalkinshaw:
I'm confused about the CI issue since there was a CI run and it passed?

I'm going to test this out later but PR looks good to me.

TangRufus:
Weird, my first commit didn't triggered CircleCI but the last 2 did.
Ignore me.

TangRufus:
Added integration tests.

swalkinshaw:
That's awesome. My only complaint is having the submodules by default. What if part of CI (and manual instructions) just said to clone them? Maybe I'm concerned over nothing though

TangRufus:
Using docker to run integration tests instead of git submodule.

Todo: 
- [ ] Push `rootsdev/trellis-cli-dev` to docker hub
- [ ] Change docker image in `.circleci/connfig.yml`

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Ansible
+# https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#latest-releases-via-apt-debian
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 93C4A3FD7BB9C367 && \
+    echo "deb http://ppa.launchpad.net/ansible/ansible-2.7/ubuntu trusty main" | tee -a /etc/apt/sources.list && \
+    apt-get -q update && \
+    apt-get install -q -y --no-install-recommends ansible && \
+    apt-get clean && apt-get -y autoremove && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Trellis
```
swalkinshaw: Maybe in the future we could use trellis-cli itself to create a new project which takes care of this (and would test trellis-cli itself)!

PR #52, opened 2019-08-12
Add `alias` command

TangRufus:
Generate WP CLI aliases for remote environments. See: https://github.com/ItinerisLtd/cognomen/issues/3#issuecomment-456563138

Ported from https://github.com/ItinerisLtd/cognomen/tree/88f77e416da8cb6bdb48deee430b9ccf4a7a486e

### DIFF
--- a/cmd/alias.go
+++ b/cmd/alias.go
@@ -1,0 +1,155 @@
+package cmd
+
+import (
+	"flag"
+	"fmt"
+	"github.com/fatih/color"
+	"github.com/mitchellh/cli"
+	"github.com/posener/complete"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	. "trellis-cli/templates"
+	"trellis-cli/trellis"
+)
+
+type AliasCommand struct {
+	UI      cli.Ui
+	flags   *flag.FlagSet
+	Trellis *trellis.Trellis
+	local   string
+}
+
+func NewAliasCommand(ui cli.Ui, trellis *trellis.Trellis) *AliasCommand {
+	c := &AliasCommand{UI: ui, Trellis: trellis}
+	c.init()
+	return c
+}
+
+func (c *AliasCommand) init() {
+	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
+	c.flags.Usage = func() { c.UI.Info(c.Help()) }
+	c.flags.StringVar(&c.local, "local", "development", "local environment name, default: development")
+}
+
+func (c *AliasCommand) Run(args []string) int {
+	if err := c.Trellis.LoadProject(); err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	if err := c.flags.Parse(args); err != nil {
+		return 1
+	}
+
+	args = c.flags.Args()
+
+	switch len(args) {
+	case 0:
+	default:
+		c.UI.Error(fmt.Sprintf("Error: too many arguments (expected 0, got %d)\n", len(args)))
+		c.UI.Output(c.Help())
+		return 1
+	}
+
+	// Template playbook files to Trellis
+	files := map[string]string{
+		"alias.yml":      ALIAS_YML,
+		"alias.yml.j2":   ALIAS_YML_J2,
+		"alias-copy.yml": ALIAS_COPY_YML,
+	}
+	for filename, fileContent := range files {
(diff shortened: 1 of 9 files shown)


cmd/alias.go
```diff
@@ -0,0 +1,153 @@
...
+
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error running ansible-playbook alias.yml: %s", err))
+			return 1
+		}
+	}
+
+	combined := ""
+	for _, environment := range remoteEnvironments {
+		part, err := ioutil.ReadFile(tempDir + "/" + environment + ".yml.part")
```
swalkinshaw: It might matter for Windows (assuming we support it properly), but we should probably always use `filepath.Join` for proper separators

cmd/alias.go
```diff
@@ -0,0 +1,153 @@
...
+		if err != nil {
+			log.Fatal(err)
+		}
+		combined = combined + string(part)
+	}
+
+	writeFileErr := ioutil.WriteFile(tempDir + "/combined.yml", []byte(combined), 0644)
```
swalkinshaw: ditto here too

cmd/alias.go
```diff
@@ -0,0 +1,153 @@
...
+      --local (default: development) Local environment name
+  -h, --help  show this help
+`
+
+	return strings.TrimSpace(helpText)
+}
+
+func (c *AliasCommand) AutocompleteFlags() complete.Flags {
+	return complete.Flags{
+		"--local": complete.PredictNothing,
```
swalkinshaw: We could autocomplete environment names. `c.Trellis.AutocompleteEnvironment()` may just work here instead

cmd/alias.go
TangRufus: good catch

cmd/alias.go
TangRufus: good catch

cmd/alias.go
TangRufus: Added in nhttps://github.com/roots/trellis-cli/pull/52/commits/e49ea2126c7062a1006060730c9d737bf7c6becf, but not sure how to verify it is workinng

swalkinshaw: I test it out locally by adding the build binary to `$PATH` and manually installing the shell completions (not on that laptop so I forget the cmd, but I also use zsh).

TangRufus: Added it into my `$PATH`

```
➜ trellis-cli --autocomplete-install
2019/08/20 17:45:14 1 error occurred:
	* already installed in /Users/me/.zshrc
```

Verified `.zshrc` has:

```
complete -o nospace -C /Users/tangrufus/Code/trellis-cli/bin/trellis-cli trellis
```

However, no shell completions avaiable for me.
```
➜ trellis-cli alias --local=
```

Note: There could be a chance that my local zsh setup is incorrect.

swalkinshaw: And that's from within a valid Trellis project? Oh you know what... that completion may only work for positional args. I don't think any other commands use it for flags yet so that's my guess

TangRufus: Reverted to `PredictNothing`.

However, we found another issue: `runtime.Caller` doesn't work when installing via `brew install --HEAD XXX/tap/trellis-cli-dev`. To be investigated.

TangRufus:
Turns out go binary doesn't include non `.go` files, i.e: `playbooks/xxx.yml` files are not complied into the binary.

When installing via `brew install --HEAD XXX/tap/trellis-cli-dev`:
1. homebrew clone the repo into a temporary directory
2. homebrew complies (`go build`) within that temporary directory
3. when invoking `trellis alias`, https://github.com/roots/trellis-cli/blob/01a20bdb71aab33c19901e1c6bb36edb5ec478e7/cmd/alias.go#L63 is referencing to the temporary directory
4. it fails because temporary directory might be delete by now

There are few alertantives to handle this problem, see: 
- https://blog.golang.org/generate
- https://github.com/shurcooL/vfsgen#alternatives

Not sure why it works when installing via goreleaser artifacts (stable versions).

Question: Anyone know why the stable versions work?

swalkinshaw:
🤔 good question. Dumb question, but are you sure it works? (did you test with the `dotenv` cmd?) Just wondering if maybe because we're building these binaries locally something in your local env made it work.

Because whenever I search anything to do with bundling static files people always mention using packages to enable it.

PR #56, opened 2019-08-22
[WIP] Homebrew: Allow install from HEAD

TangRufus:
https://github.com/TangRufus/homebrew-tap/blob/feac3f9776e86d7e65910b7a5cafd00e30b56e76/Formula/trellis-cli.rb works, but not sure how to test goreleaser generates the formula as expected.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -38,8 +38,16 @@ brews:
     folder: Formula
     homepage:  https://roots.io/trellis
     description: A CLI to manage Trellis projects
+    custom_block: |
+      head "https://github.com/roots/trellis-cli.git"
+      depends_on "go" => :build
     install: |
-      bin.install "trellis"
+      output = "trellis"
+      if build.head?
+        output = "bin/" + output
+        system "go", "build", "-o", output
+      end
+      bin.install output
     test: |
       system "#{bin}/trellis --autocomplete-install"
       system "#{bin}/trellis -v"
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -38,8 +38,16 @@ brews:
     folder: Formula
     homepage:  https://roots.io/trellis
     description: A CLI to manage Trellis projects
+    custom_block: |
+      head "https://github.com/roots/trellis-cli.git"
+      depends_on "go" => :build
```
swalkinshaw: As long as this means it's only needed for build/head 👍

TangRufus: Good catch!
Seems a goreleaser tap always install build dependencyies.

TangRufus: Seems goreleaser always install `depends_on "go" => :build`.

What do you think if we make another formula (trellis-cli-dev) in https://github.com/roots/homebrew-tap

Something like what sequel does:
```sh-session
$ brew search sequel
==> Casks
psequel                    sequel-pro                 sequel-pro-nightly ✔
```

swalkinshaw: I'm good with that 👍 I think that just means adding another brew entry in the goreleaser config in this repo?

TangRufus: I think https://github.com/roots/homebrew-tap/pull/1 is enough because that formula doesn't need to be changed by goreleaser

swalkinshaw:
`brew install roots/trellis-cli-dev` should work now

PR #61, opened 2019-09-04
Rename Ansible role file for `galaxy install` cmd

knowler:
From `requirements.yml` to `galaxy.yml` as per Trellis v1.1.0 (specifically roots/trellis#1100).

It’s probably preferable to have it set the file name based on the Trellis version, since this change alone would break compatibility with Trellis versions prior to 1.1.0.

### DIFF
--- a/cmd/galaxy_install.go
+++ b/cmd/galaxy_install.go
@@ -25,7 +25,7 @@ func (c *GalaxyInstallCommand) Run(args []string) int {
 		return 1
 	}
 
-	galaxyInstall := execCommand("ansible-galaxy", "install", "-r", "requirements.yml")
+	galaxyInstall := execCommand("ansible-galaxy", "install", "-r", "galaxy.yml")
 	logCmd(galaxyInstall, c.UI, true)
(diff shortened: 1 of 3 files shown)


swalkinshaw:
🤦‍♂ I made the change for better CLI integration and of course forgot to update it here. I'll probably make this smarter based on file existence, but this works for now 👍

cmd/galaxy_install.go
```diff
@@ -25,7 +25,7 @@ func (c *GalaxyInstallCommand) Run(args []string) int {
 		return 1
 	}
 
-	galaxyInstall := execCommand("ansible-galaxy", "install", "-r", "requirements.yml")
+	galaxyInstall := execCommand("ansible-galaxy", "install", "-r", "galaxy.yml")
```
TangRufus: @swalkinshaw  @knowler 

What do you think if we check for `galaxy.yml` existence and fallback to `requirements.yml` for older versions of Trellis?

swalkinshaw: yep that's what in my other comment 👍

PR #62, opened 2019-09-05
Galaxy install: handle both role files

swalkinshaw:
Adds backwards compatibility for `galaxy install` to support either
`requirements.yml` or `galaxy.yml`

### DIFF
--- a/cmd/galaxy_install_test.go
+++ b/cmd/galaxy_install_test.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"os"
 	"os/exec"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


cmd/galaxy_install_test.go
```diff
@@ -89,5 +119,9 @@ func TestGalaxyInstallRun(t *testing.T) {
 		if !strings.Contains(combined, tc.out) {
 			t.Errorf("expected output %q to contain %q", combined, tc.out)
 		}
+
+		for _, file := range tc.roleFiles {
+			os.Remove(file)
```
TangRufus: How about move this to immediately after line 108 and use `defer`?

swalkinshaw: I had that originally but `defer` only runs before exit and this needs to be removed on *each* loop iteration.

TangRufus: My bad

PR #68, opened 2019-09-23
Remove windows support

TangRufus:
This is because `syscall.Exec` doesn't work on windows and `sys/unix.Exec` doesn't have a windows counterpart. See: https://github.com/roots/trellis-cli/pull/64

Until someone smarter than me make [`SyscallCommandExecutor`](https://github.com/roots/trellis-cli/blob/cb5abb0c32042c67567f476f22fb7afbbff4f89a/cmd/command.go#L16) works on windows, use WSL instead. 

Sorry windows users.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -8,7 +8,6 @@ builds:
   goos:
     - linux
     - darwin
-    - windows
   goarch:
     - 386
     - amd64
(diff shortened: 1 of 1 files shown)


austinpray:
RIP in pepperonis

TangRufus:
Not sure whether we need to regenerate https://github.com/roots/trellis-cli/blob/cb5abb0c32042c67567f476f22fb7afbbff4f89a/scripts/get

PR #72, opened 2019-09-25
Add `db open` command

TangRufus:
Open remote databases via SSH tunnels in your favourite database client.

### Requirement

An database client on your local machine registered as `mysql+ssh://` handler, e.g: [TablePlus](https://tableplus.com/)

### How does it work?

In a nutshell, it invokes this command on your local machine:

```
$ open 'mysql+ssh://web@111.222.333.444:22/my_db_user:my_db_password@localhost/my_db_name?enviroment=production&usePrivateKey=true'
```

### Which database clients are supported?

I've only tested [TablePlus](https://tableplus.com/).

Sqeual Pro (nightly) supports `mysql://` but seems [not support SSH tunnels yet](https://github.com/sequelpro/sequelpro/issues/1681)

### Does it support vagrant / local databases?

No.

---

See: https://github.com/roots/trellis-cli/issues/7

### DIFF
--- a/cmd/db_open.go
+++ b/cmd/db_open.go
@@ -166,7 +166,7 @@ func (c *DBOpenCommand) Synopsis() string {
 
 func (c *DBOpenCommand) Help() string {
 	helpText := `
-Usage: trellis deploy [options] ENVIRONMENT [SITE]
+Usage: trellis db [options] ENVIRONMENT [SITE]
 
 Open database with GUI applications
 
--- a/cmd/db_open.go
+++ b/cmd/db_open.go
@@ -12,7 +12,8 @@ import (
 	"trellis-cli/trellis"
 )
 
-var dumpDbCredentialsYmlTemplate = `
+const playbookPath = "dump_db_credentials.yml"
+const dumpDbCredentialsYmlTemplate = `
 ---
 - name: 'Trellis CLI: Dump database credentials'
   hosts: web:&{{ env }}
@@ -29,7 +30,8 @@ var dumpDbCredentialsYmlTemplate = `
       when: item.key == site
 `
 
-var dbCredentialsJsonJ2Template = `
+const j2TemplatePath = "db_credentials.json.j2"
+const dbCredentialsJsonJ2Template = `
 {
     "web_user": "{{ web_user }}",
     "ansible_host": "{{ ansible_host }}",
@@ -114,10 +116,8 @@ func (c *DBOpenCommand) Run(args []string) int {
 	defer deleteFile(dbCredentialsJson.Name())
 
 	// Template playbook files from package to Trellis
-	playbookPath := "dump_db_credentials.yml"
 	writeFile(playbookPath, templates.TrimSpace(dumpDbCredentialsYmlTemplate))
 	defer deleteFile(playbookPath)
-	j2TemplatePath := "db_credentials.json.j2"
 	writeFile(j2TemplatePath, templates.TrimSpace(dbCredentialsJsonJ2Template))
 	defer deleteFile(j2TemplatePath)
 
--- a/cmd/db_open.go
+++ b/cmd/db_open.go
@@ -1,0 +1,187 @@
+package cmd
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/mitchellh/cli"
+	"trellis-cli/templates"
+	"trellis-cli/trellis"
+)
+
+var dumpDbCredentialsYmlTemplate = `
+---
+- name: 'Trellis CLI: Dump database credentials'
+  hosts: web:&{{ env }}
+  remote_user: "{{ web_user }}"
+  gather_facts: false
+  connection: local
+  tasks:
+    - name: Dump database credentials
+      template:
+        src: db_credentials.json.j2
+        dest: "{{ dest }}"
+        mode: '0600'
+      with_dict: "{{ wordpress_sites }}"
+      when: item.key == site
+`
+
+var dbCredentialsJsonJ2Template = `
+{
+    "web_user": "{{ web_user }}",
+    "ansible_host": "{{ ansible_host }}",
+    "ansible_port": {{ ansible_port | default(22) }},
+    "db_user": "{{ site_env.db_user }}",
+    "db_password": "{{ site_env.db_password }}",
+    "db_host": "{{ site_env.db_host }}",
+    "db_name": "{{ site_env.db_name }}",
+    "wp_env": "{{ site_env.wp_env }}"
+}
+`
+
+func NewDBOpenCommand(ui cli.Ui, trellis *trellis.Trellis, dbOpenerFactory *DBOpenerFactory) *DBOpenCommand {
+	c := &DBOpenCommand{UI: ui, Trellis: trellis, dbOpenerFactory: dbOpenerFactory}
+	c.init()
+	return c
+}
+
+type DBOpenCommand struct {
+	UI              cli.Ui
+	flags           *flag.FlagSet
+	app             string
+	Trellis         *trellis.Trellis
+	dbOpenerFactory *DBOpenerFactory
+}
+
+type DBCredentials struct {
+	SSHUser    string `json:"web_user"`
+	SSHHost    string `json:"ansible_host"`
+	SSHPort    int    `json:"ansible_port"`
+	DBUser     string `json:"db_user"`
+	DBPassword string `json:"db_password"`
+	DBHost     string `json:"db_host"`
+	DBName     string `json:"db_name"`
+	WPEnv      string `json:"wp_env"`
+}
+
+func (c *DBOpenCommand) init() {
+	c.flags = flag.NewFlagSet("", flag.ContinueOnError)
+	c.flags.Usage = func() { c.UI.Info(c.Help()) }
+	c.flags.StringVar(&c.app, "app", "", "Database client to be used; Supported: tableplus, sequel-pro")
+}
+
+func (c *DBOpenCommand) Run(args []string) int {
+	if err := c.Trellis.LoadProject(); err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	if flagsParseErr := c.flags.Parse(args); flagsParseErr != nil {
+		return 1
+	}
+
+	args = c.flags.Args()
+
+	commandArgumentValidator := &CommandArgumentValidator{required: 1, optional: 1}
+	if commandArgumentErr := commandArgumentValidator.validate(args); commandArgumentErr != nil {
+		c.UI.Error(commandArgumentErr.Error())
+		c.UI.Output(c.Help())
+		return 1
+	}
+
+	environment := c.flags.Arg(0)
+	environmentErr := c.Trellis.ValidateEnvironment(environment)
+	if environmentErr != nil {
+		c.UI.Error(environmentErr.Error())
+		return 1
+	}
+
+	siteNameArg := c.flags.Arg(1)
+	siteName, siteNameErr := c.Trellis.FindSiteNameFromEnvironment(environment, siteNameArg)
+	if siteNameErr != nil {
+		c.UI.Error(siteNameErr.Error())
+		return 1
+	}
+
+	// Template JSON file for db credentials
+	dbCredentialsJson, dbCredentialsErr := ioutil.TempFile("", "*.json")
+	if dbCredentialsErr != nil {
+		c.UI.Error(fmt.Sprintf("Error createing temporary db credentials JSON file: %s", dbCredentialsErr))
+	}
+	defer deleteFile(dbCredentialsJson.Name())
+
+	// Template playbook files from package to Trellis
+	playbookPath := "dump_db_credentials.yml"
+	writeFile(playbookPath, templates.TrimSpace(dumpDbCredentialsYmlTemplate))
+	defer deleteFile(playbookPath)
+	j2TemplatePath := "db_credentials.json.j2"
+	writeFile(j2TemplatePath, templates.TrimSpace(dbCredentialsJsonJ2Template))
+	defer deleteFile(j2TemplatePath)
+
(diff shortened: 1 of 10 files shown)


TangRufus:
Discussion:

How should we handle app-specific paramters (e.g: `?enviroment={{ site_env.wp_env }}&usePrivateKey=true`)?

ping @swalkinshaw

templates/db_open_yml.go
```diff
@@ -0,0 +1,15 @@
...
+const DB_OPEN_YML = `
+---
+- name: 'Trellis CLI: Open database connection'
+  hosts: web:&{{ env }}
+  remote_user: "{{ web_user }}"
+  gather_facts: false
+  connection: local
+  tasks:
+    - name: Open database connection
+      shell: "open 'mysql+ssh://{{ web_user }}@{{ ansible_host }}:{{ ansible_port | default('22') }}/{{ site_env.db_user }}:{{ site_env.db_password }}@{{ site_env.db_host }}/{{ site_env.db_name }}?enviroment={{ site_env.wp_env }}&usePrivateKey=true'"
```
swalkinshaw: 🤔 Wonder if there's an easy to have this playbook just dump out the variables we need as JSON/YAML. Then trellis-cli could parse it and handle actually executing the shell command. I think it's a bit better to keep these "embedded" playbooks as dumb as possible. What do you think?

swalkinshaw:
> How should we handle app-specific paramters

By app I'm guessing you mean TablePlus and Sequel Pro for example? I guess we'd have to have some we officially support and hardcode their necessary options? Then we could also let people pass-through whatever extra params they need.

templates/db_open_yml.go
TangRufus: It is possible. Similar to the `dotenv` command.

I guess we are going to dump them in `$TMP` (or `TMP_DIR`).
Do we need to worry about they are not being deleted if `$ trellis db open` half way through?
(note the dump contains db password)

TangRufus:
> By app I'm guessing you mean TablePlus and Sequel Pro for example? 

Yes.

What do you think about this style?

```
trellis db open production example.com tableplus
trellis db open production example.com sequel-pro 
```

The reason is that they can be open in very different ways:

TablePlus: `$ open 'mysql+ssh://web@111.222.333.444:22/my_db_user:my_db_password@localhost/my_db_name?enviroment=production&usePrivateKey=true'`

Sequal Pro: `$ open  xxx.spf`
(see: https://github.com/TypistTech/vagrant-trellis-sequel/blob/cc93696de180ba8e72e80039c9837f0de26bd77c/lib/vagrant_plugins/trellis_sequel/template.spf.erb)

templates/db_open_yml.go
swalkinshaw: Oh I was just assuming we'd get the output from STDOUT and parse it so there wouldn't be a need for any temp files.

swalkinshaw:
We don't really have other options than your suggestion but I think they should be option flags instead?

templates/db_open_yml.go
TangRufus: I worry getting from STDOUT would be troublesome.
Ansible output is not easy to parse and ansible updates might break the parser

TangRufus:
#### Help wanted

https://github.com/roots/trellis-cli/pull/72/commits/10ad118e2eb535090da6c0775a70b1d460e2f228 introduces `$ trellis db open --app=tableplus production mysite`. However, the implementation is so bad that writting tests is painful.

Whoever can't stand with my code smell, free feel to make pull requests to [my branch](https://github.com/TangRufus/trellis-cli/tree/db-open).

---

> assuming we'd get the output from STDOUT and parse it so there wouldn't be a need for any temp files.
> https://github.com/roots/trellis-cli/pull/72#discussion_r328856141

I have trouble getting ansible outputs from `STDOUT`.

---

There are lots of boilerplate code . Example: 
- https://github.com/roots/trellis-cli/pull/72/files#diff-a08b51a976e893b57653c908b219ffbcR76-R114
- https://github.com/roots/trellis-cli/pull/72/commits/10ad118e2eb535090da6c0775a70b1d460e2f228#diff-a08b51a976e893b57653c908b219ffbcR124-R141

However, refactoring them affects almost every existing commands.
Perphas leave it to another pull request.

---

Bash autocompletion doesn't work on my machine. Thus, omitted.

---

Tried to put `cmd/tableplus.go` into subdirectory - `cmd/db/openers/tableplus.go`.
However, [`execCommand` & `logCmd`](https://github.com/roots/trellis-cli/blob/c043cfd8704fda69dee5387a054ae2be88e1fc1c/cmd/main.go) become unavaible. 

Same goes to `cmd/db/openers/factory.go` (https://github.com/roots/trellis-cli/pull/72/files#diff-a08b51a976e893b57653c908b219ffbcR44-R53)

swalkinshaw:
🤷‍♂ I think it looks mostly good overall. Not sure we can really avoid a lot of boilerplate here (and it's Go after all 😛 ).

re: subdirectory, Go is weird about this and I don't know the best solution. But for now I'd just name the files `cmd/db_opener_tableplus.go` etc.

TangRufus:
Ignore this PR until https://github.com/roots/trellis-cli/pull/88 is merged

TangRufus:
Added more tests.

Note:
- `DBOpenCommand` is untested (Help wanted)
- Not sure how to test `DBOpenerFactory` returns correct strcut (Any golang  equivalent of `instanceof`?)
- `DBOpenerSequelPro` is not removing its spf file in temporary directory because the spf file might be removed before sequel pro reads db credentials from it. (Warning: The SPF contains credentials in plain text)
- Intentionally omitting `logCmd` from `DBOpenerTableplus` to prevent printing db credentials.

swalkinshaw:
For testing the struct type, I think you want https://golang.org/pkg/reflect/#TypeOf

cmd/db_open.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// Template JSON file for db credentials
+	dbCredentialsJson, dbCredentialsErr := ioutil.TempFile("", "*.json")
+	if dbCredentialsErr != nil {
+		c.UI.Error(fmt.Sprintf("Error createing temporary db credentials JSON file: %s", dbCredentialsErr))
+	}
+	defer deleteFile(dbCredentialsJson.Name())
+
+	// Template playbook files from package to Trellis
+	playbookPath := "dump_db_credentials.yml"
```
swalkinshaw: Maybe make this a `const` up top?

cmd/db_open.go
```diff
@@ -0,0 +1,187 @@
...
+	writeFile(playbookPath, templates.TrimSpace(dumpDbCredentialsYmlTemplate))
+	defer deleteFile(playbookPath)
+	j2TemplatePath := "db_credentials.json.j2"
```
swalkinshaw: This as well

cmd/db_open.go
```diff
@@ -0,0 +1,187 @@
...
+	// Template JSON file for db credentials
+	dbCredentialsJson, dbCredentialsErr := ioutil.TempFile("", "*.json")
+	if dbCredentialsErr != nil {
+		c.UI.Error(fmt.Sprintf("Error createing temporary db credentials JSON file: %s", dbCredentialsErr))
+	}
+	defer deleteFile(dbCredentialsJson.Name())
+
+	// Template playbook files from package to Trellis
+	playbookPath := "dump_db_credentials.yml"
+	writeFile(playbookPath, templates.TrimSpace(dumpDbCredentialsYmlTemplate))
```
swalkinshaw: Every usage of `writeFile` deletes the file after. Maybe we should just make this a more specific `writeTempFile` function which also handles the defer delete?

cmd/db_open.go
```diff
@@ -0,0 +1,187 @@
...
+	return 0
+}
+
+func (c *DBOpenCommand) Synopsis() string {
+	return "Open database with GUI applications"
+}
+
+func (c *DBOpenCommand) Help() string {
+	helpText := `
+Usage: trellis deploy [options] ENVIRONMENT [SITE]
```
swalkinshaw: This needs updating to be `db`

cmd/db_opener_sequel_pro.go
```diff
@@ -0,0 +1,84 @@
...
+</dict>
+</plist>
+`
+
+func (o *DBOpenerSequelPro) open(c DBCredentials) (err error) {
+	sequelProSpf, sequelProSpfErr := ioutil.TempFile("", "*.spf")
+	if sequelProSpfErr != nil {
+		return fmt.Errorf("Error creating temporary SequelPro SPF file: %s", sequelProSpfErr)
+	}
+	// TODO: [Help Wanted] There is a chance that the SPF file got deleted before SequelPro establish db connection.
```
swalkinshaw: Can you explain this more? Was it getting deleted by the `defer` first?

swalkinshaw:
I can work on some tests as well if you want

cmd/db_open.go
TangRufus: done

cmd/db_open.go
TangRufus: done

cmd/db_open.go
TangRufus: done

TangRufus:
> For testing the struct type, I think you want golang.org/pkg/reflect/#TypeOf

It works!

> There's already some tests for DBOpenerSequelPro; wouldn't those work the same way for DBOpenCommand as well? Just assert on the mock exec command?

Not sure how to mock the json file - https://github.com/roots/trellis-cli/pull/72/files#diff-a08b51a976e893b57653c908b219ffbcR135

> I can work on some tests as well if you want

Please do! Thanks!

cmd/db_open.go
TangRufus: How do we defer the `defer`?

If we put `defer` inside `writeTempFile`, it deletes teh files at the end of `writeTempFile` instead of the end of `func (c *DBOpenCommand) Run(args []string) int`

cmd/db_opener_sequel_pro.go
TangRufus: > Was it getting deleted by the defer first?

That is my theory.

If we uncomment the `defer` line, sequel pro will open but doesn't pre-fill any db info.
If we comment the `defer` line, sequel pro will open with db info pre-filled and try to connect to the db.

I suspect sequel pro needs to read that spf file after launch, but golang deleted the spf too soon.

cmd/db_open.go
swalkinshaw: Oh yes good call. We do this: https://github.com/roots/trellis-cli/blob/54aeb5d6fdf9df75afdb59c48fbca34d64fd2125/trellis/testing.go#L32-L38

Then it would be called as:

```go
defer writeFile(path, foo)()
```

cmd/db_opener_sequel_pro.go
swalkinshaw: Makes some sense. You could verify by logging like this:

```go
func (o *DBOpenerSequelPro) open(c DBCredentials) (err error) {
         # etc
	open := execCommand("open", sequelProSpf.Name())
	err := cmd.Start()
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("running sequel pro...")
	err = cmd.Wait()
	log.Printf("Command finished with error: %v", err)
}
```

And then add a log printf to `deleteFile` to see when it's actually run compared to the other log lines

swalkinshaw:
I looked into testing and it's going to be very hard... it *is* possible to have more control over the mock exec command but it's not nice. We'd have to hardcode a conditional in `TestHelperProcess` to do what we want for the first ansible command. But that doesn't even solve the issue with writing to a tempfile that we wouldn't know anything about.

I think there's two solutions:
1. write an interface which models writing the temp file, running ansible cmd and reading the temp file then we could make a test implementation of the struct
2. make this an integration test that runs on Docker

TangRufus:
> write an interface which models writing the temp file, running ansible cmd and reading the temp file then we could make a test implementation of the struct

Something like https://github.com/TangRufus/trellis-cli/compare/b83f85e...TangRufus:playbook-struct ?

swalkinshaw:
Yep that looks like it should do it 👍 . The test struct would mostly be concerned with implementing `Run` and would just dump some fixture JSON to the tempfile.

TangRufus:
I assume that we want to inject `Playbook` in [`main.go`](https://github.com/roots/trellis-cli/blob/b83f85eb5df9fdee66c0b9e4a03b1513061ab673/main.go#L26-L34).

```
 	trellis := trellis.NewTrellis(project)
+   playbook := &Playbook{}
	 c.Commands = map[string]cli.CommandFactory{
		 "alias": func() (cli.Command, error) {
-			 return cmd.NewAliasCommand(ui, trellis), nil
+			 return cmd.NewAliasCommand(ui, trellis, playbook), nil
		 },
```

Am I correct?

swalkinshaw:
Correct yeah. That's the only place available and you can see examples in `ssh` and `vault edit` for the sys command executor.

TangRufus:
Question: What are the differences between `exec.Command` and `unix.Exec`?

swalkinshaw:
The way I understand it:

* `os/exec`: the go process is still "managing" the new command/process
* `syscall` (or `unix.Exec`): runs a new process which completely takes over from the current one

I found it was needed for `ssh` for that process to take over and with `vault edit` it launches an editor. Maybe there is a way to do that with `os/exec` but I couldn't get it working.

TangRufus:
The merge was a mistake. Sorry everyone. Reverted in https://github.com/roots/trellis-cli/pull/90

cmd/db_opener_sequel_pro.go
TangRufus: Confirmed my theroy - the SPF file is read twice:
- once by OS to open sequel pro
- once by sequel pro to get db credentials

---

Newer version of sequel pro shows this error message if SPF is deleted too early:

```
Error while reading connection data file

Connection data file couldn't be read. (The file “804911772.spf” couldn’t be opened because there is no such file.)
```

PR #80, opened 2019-10-08
Use fmt.Errorf to create formatted error string

sjothen:
It's just a bit more succinct, eh?


swalkinshaw:
Oh nice I thought that just returned a string for some reason 👍

swalkinshaw:
There might be other occurrences of this

PR #82, opened 2019-10-09
Remove unreachable file existence check

sjothen:
Stat returns nil for error if file exists, so that check will never run, removing it keeps the same functionality.

### DIFF
--- a/trellis/trellis.go
+++ b/trellis/trellis.go
@@ -38,10 +38,6 @@ func (t *Trellis) Detect(path string) (projectPath string, ok bool) {
 func (t *Trellis) CreateConfigDir() error {
 	_, err := os.Stat(t.ConfigPath)
 
-	if os.IsExist(err) {
-		return nil
-	}
-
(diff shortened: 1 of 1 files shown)


trellis/trellis.go
```diff
@@ -38,10 +38,6 @@ func (t *Trellis) Detect(path string) (projectPath string, ok bool) {
 func (t *Trellis) CreateConfigDir() error {
 	_, err := os.Stat(t.ConfigPath)
```
swalkinshaw: Maybe this should just try to always to `Mkdir` and then check if the error is `IsExist`? If it's not, return the error. Something like

```go
if err := os.Mkdir(t.ConfigPath, 0755); err != nil {
  if !os.IsExist(err) {
    return err
  }
}

return nil
```

sjothen: Good idea, one less syscall, super optimized!

swalkinshaw:
Woot 🎉

PR #86, opened 2019-11-03
Extract `validateArgumentCount`

TangRufus:
- simplify argument count validation logic
- normalize argument count error message

---

Example usage:

```golang
# $ trellis deploy [options] ENVIRONMENT [SITE]

args = c.flags.Args()

argCountErr := validateArgumentCount(args, 1, 1)
if argCountErr != nil {
	c.UI.Error(argCountErr.Error())
	c.UI.Output(c.Help())
	return 1
}

environment := args[0]

siteName := ""
if len(args) == 2 {
	siteName = args[1]
}
```

---

Note: `validateArgumentCount` doesn't support unlimited arguments, i.e: https://github.com/roots/trellis-cli/blob/c043cfd8704fda69dee5387a054ae2be88e1fc1c/cmd/exec.go#L28-L36

### DIFF
--- a/cmd/alias.go
+++ b/cmd/alias.go
@@ -46,10 +46,9 @@ func (c *AliasCommand) Run(args []string) int {
 
 	args = c.flags.Args()
 
-	switch len(args) {
-	case 0:
-	default:
-		c.UI.Error(fmt.Sprintf("Error: too many arguments (expected 0, got %d)\n", len(args)))
+	argCountErr := validateArgumentCount(args, 0, 0)
+	if argCountErr != nil {
+		c.UI.Error(argCountErr.Error())
 		c.UI.Output(c.Help())
 		return 1
 	}
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -33,7 +33,9 @@ var Requirements = []trellis.Requirement{
 }
 
 func (c *CheckCommand) Run(args []string) int {
-	if len(args) > 0 {
+	argCountErr := validateArgumentCount(args, 0, 0)
+	if argCountErr != nil {
+		c.UI.Error(argCountErr.Error())
 		c.UI.Output(c.Help())
 		return 1
 	}
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -50,6 +50,7 @@ func (c *DeployCommand) Run(args []string) int {
 	}
 
 	environment := args[0]
+
 	siteName := ""
 	if len(args) == 2 {
 		siteName = args[1]
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -36,28 +36,23 @@ func (c *DeployCommand) Run(args []string) int {
 		return 1
 	}
 
-	var environment string
-	var siteName string
-
 	if err := c.flags.Parse(args); err != nil {
 		return 1
 	}
 
 	args = c.flags.Args()
 
-	switch len(args) {
-	case 0:
+	argCountErr := validateArgumentCount(args, 1, 1)
+	if argCountErr != nil {
+		c.UI.Error(argCountErr.Error())
 		c.UI.Output(c.Help())
 		return 1
-	case 1:
-		environment = args[0]
-	case 2:
-		environment = args[0]
+	}
+
+	environment := args[0]
+	siteName := ""
+	if len(args) == 2 {
 		siteName = args[1]
-	default:
-		c.UI.Error(fmt.Sprintf("Error: too many arguments (expected 2, got %d)\n", len(args)))
-		c.UI.Output(c.Help())
-		return 1
 	}
 
 	_, ok := c.Trellis.Environments[environment]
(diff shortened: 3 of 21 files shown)


cmd/alias.go
```diff
@@ -46,10 +46,9 @@ func (c *AliasCommand) Run(args []string) int {
 
 	args = c.flags.Args()
 
-	switch len(args) {
-	case 0:
-	default:
-		c.UI.Error(fmt.Sprintf("Error: too many arguments (expected 0, got %d)\n", len(args)))
+	argCountErr := validateArgumentCount(args, 0, 0)
+	if argCountErr != nil {
```
swalkinshaw: What if we made this a struct with methods too? It could look something like:

```go
type CommandArgumentValidator struct {
  args []string{}
  required Int
  optional Int
  // or maybe initialize with UI too
}

func (c *CommandArgumentValidator) validate(ui cli.Ui) int {
  # etc
  ui.Error(argCountErr.Error())
  ui.Output(c.Help())
  return 1
}

// in a command

validator := CommandArgumentValidator{args: args, required: 0, optional: 0}
validator.validate(c.UI)
```

One reason I like this is for the named parameters

TangRufus: Updated.

Usage:

```golang
commandArgumentValidator := &CommandArgumentValidator{required: 1, optional: 1}
commandArgumentErr := commandArgumentValidator.validate(args)
if commandArgumentErr != nil {
	c.UI.Error(commandArgumentErr.Error())
	c.UI.Output(c.Help())
	return 1
}
```

---

To keep `CommandArgumentValidator` simple., I am not passing `c.UI` & `c.Help()` to `CommandArgumentValidator`.

Polluting `CommandArgumentValidator` with `c.UI` & `c.Help()` to make 2 less lines of code doesn't worth and looks like a violation of single responsibility principle to me. :
```goland
commandArgumentValidator := &CommandArgumentValidator{required: 1, optional: 1, UI: c.UI, Help: c.Help()}
commandArgumentErr := commandArgumentValidator.validate(args)
if commandArgumentErr != nil {
	return 1
}
```

What do you think?

swalkinshaw: I think I agree 👍

cmd/command_argument_validator.go
```diff
@@ -0,0 +1,25 @@
...
+	required int
+	optional int
+}
+
+func (c *CommandArgumentValidator) validate(args []string) (err error) {
+	argCount := len(args)
+
+	expectedCount := fmt.Sprintf("exactly %d", c.required)
+	if (c.optional > 0) {
+		expectedCount = fmt.Sprintf("between %d and %d", c.required, c.required+c.optional)
```
swalkinshaw: Maybe just declare `totalArgs := c.required+c.optional` up front?

cmd/command_argument_validator.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	if argCount > c.required+c.optional {
+		err = fmt.Errorf("Error: too many arguments (expected %s, got %d)\n", expectedCount, len(args))
```
swalkinshaw: You can re-use `argCount` in both places here?

cmd/command_argument_validator.go
```diff
@@ -0,0 +1,25 @@
...
+	} else if argCount < c.required {
+		err = fmt.Errorf("Error: missing arguments (expected %s, got %d)\n", expectedCount, len(args))
+	}
+
+	return
```
swalkinshaw: 🤔 does it complain if you explicitly do `return err`? Makes more sense to me if not (even though this works)

cmd/command_argument_validator.go
TangRufus: done

cmd/command_argument_validator.go
TangRufus: done

cmd/command_argument_validator.go
TangRufus: - added explicitly return in case https://github.com/golang/go/issues/21291 happens
- return inside the if block to short-circuit it

cmd/command_argument_validator.go
```diff
@@ -0,0 +1,27 @@
...
+	expectedCount := fmt.Sprintf("exactly %d", c.required)
+	if (c.optional > 0) {
+		expectedCount = fmt.Sprintf("between %d and %d", c.required, totalArgs)
+	}
+
+	if argCount > totalArgs {
+		return fmt.Errorf("Error: too many arguments (expected %s, got %d)\n", expectedCount, len(args))
+	}
+	if argCount < c.required {
+		return fmt.Errorf("Error: missing arguments (expected %s, got %d)\n", expectedCount, len(args))
```
swalkinshaw: Sorry I meant you can use `argCount` instead of `len(args)` in both these error messages?

cmd/command_argument_validator.go
TangRufus: i need coffee

cmd/command_argument_validator.go
TangRufus: i need coffee

PR #88, opened 2019-11-07
Extract `func (t *Trellis) FindSiteNameFromEnvironment(environment string, siteNameArg string) (string, error) {`

TangRufus:
- Extract `func (t *Trellis) FindSiteNameFromEnvironment(environment string, siteNameArg string) (string, error) {`
- Normalize error message

### DIFF
--- a/trellis/trellis.go
+++ b/trellis/trellis.go
@@ -129,6 +129,35 @@ func (t *Trellis) SiteNamesFromEnvironment(environment string) []string {
 	return names
 }
 
+func (t *Trellis) FindSiteNameFromEnvironment(environment string, siteNameArg string) (string, error) {
+	if siteNameArg == "" {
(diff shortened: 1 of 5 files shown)


trellis/trellis.go
```diff
@@ -129,6 +129,35 @@ func (t *Trellis) SiteNamesFromEnvironment(environment string) []string {
 	return names
 }
 
+func (t *Trellis) FindSiteNameFromEnvironment(environment string, siteNameArg string) (string, error) {
```
swalkinshaw: Should we unit test this?

TangRufus: Added along with some other `*Trellis` unit tests

swalkinshaw: Not even close 😞 but in my defense a CLI like this makes it hard sometimes

PR #91, opened 2020-01-04
Extract `Playbook` struct

TangRufus:
See: https://github.com/roots/trellis-cli/pull/72#issuecomment-555303446

---

Although all tests passed, need help on testing the commands manually to check for unwated (and not covered by tests) issues.


cmd/playbook.go
```diff
@@ -0,0 +1,77 @@
...
+	SetRoot(root string)
+	Run(playbookYml string, args []string, ui cli.Ui) error
+}
+
+type Playbook struct {
+	files map[string]string
+	root  string
+}
+
+func (p *Playbook) SetFiles(files map[string]string) {
```
swalkinshaw: 🤔 I'm not sure how I feel about `Run` optionally doing the generation of playbook templates. It's a bit too implicit and inconsistent? Since in other cases it's used to run "normal" playbooks which already exist (and won't be removed).

Couple options:
1. leave as is
2. rename `SetFiles` to be `GeneratePlaybooks` or something more explicit
3. add a new version of `Run` such as `RunTemporary` to make it more explicit

cmd/playbook.go
```diff
@@ -0,0 +1,77 @@
+package cmd
+
+import (
+	"github.com/mitchellh/cli"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+type PlaybookInterface interface {
```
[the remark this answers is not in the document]
TangRufus: done

cmd/playbook.go
TangRufus: Seperated `AdHocPlaybook` & `Playbook`

cmd/alias.go
```diff
@@ -11,19 +11,62 @@ import (
...
+        mode: '0644'
+        force: yes
+        decrypt: no
+      with_dict: "{{ wordpress_sites }}"
+      run_once: true
+`
+
 func NewAliasCommand(ui cli.Ui, trellis *trellis.Trellis) *AliasCommand {
-	c := &AliasCommand{UI: ui, Trellis: trellis}
+	c := &AliasCommand{UI: ui, Trellis: trellis, aliasPlaybook: &Playbook{}, aliasCopyPlaybook: &Playbook{}}
```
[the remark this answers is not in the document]
TangRufus: - Removed `SetFiles`
- Removed the 3rd parameter (`ui cli.Ui`) from `PlaybookRunner.Run`
Passing both during struct construction.

However, `SetRoot` has to be kept because it has to run after [`c.Trellis.LoadProject()`](https://github.com/TangRufus/trellis-cli/blob/b83f85eb5df9fdee66c0b9e4a03b1513061ab673/cmd/alias.go#L38)

swalkinshaw:
Looks great 👍 I'll try and test soon

cmd/droplet_create.go
```diff
@@ -172,10 +173,11 @@ func (c *DropletCreateCommand) Run(args []string) int {
 		galaxyInstallCmd.Run([]string{})
 
 		c.UI.Info("\nProvisioning server...\n")
-		playbookCmd := ProvisionCmd(environment, "", "")
-		logCmd(playbookCmd, c.UI, true)
 
-		if err := playbookCmd.Run(); err != nil {
+		c.playbook.SetRoot(c.Trellis.Path)
+
+		if err := c.playbook.Run("server.yml", []string{"-e", "env" + environment}); err != nil {
```
swalkinshaw: `env` is missing `=`

swalkinshaw: The Playbook struct is nice because we can easily re-use it in places like this, but I still wonder if we should actually call the `ProvisionCommand` itself here to ensure they're consistent?

```go
provisionCmd := NewProvisionCommand(c.UI, c.Trellis)
provisionCmd.Run([]string{environment})
```

Something like that?

cmd/deploy.go
```diff
@@ -75,12 +76,9 @@ func (c *DeployCommand) Run(args []string) int {
...
 	extraVars := fmt.Sprintf("\"%s\"", strings.Join(vars, " "))
 
-	playbookArgs := []string{"deploy.yml", "-e", extraVars}
-	deploy := execCommand("ansible-playbook", playbookArgs...)
-	logCmd(deploy, c.UI, true)
-	err := deploy.Run()
+	c.playbook.SetRoot(c.Trellis.Path)
 
-	if err != nil {
+	if err := c.playbook.Run("deploy.yml", []string{"-e", extraVars}); err != nil {
```
swalkinshaw: This breaks `extraVars` somehow 🤔 Ansible doesn't recognize the actual vars within `-e`

swalkinshaw: Actually it might have been broken before somehow? Trying to figure it out

cmd/deploy.go
```diff
@@ -75,12 +76,9 @@ func (c *DeployCommand) Run(args []string) int {
 
 	extraVars := fmt.Sprintf("\"%s\"", strings.Join(vars, " "))
```
swalkinshaw: Looks like this line is the problem... the first part isn't even necessary?

```go
extraVars := strings.Join(vars, " ")
```

Not sure what led me to that originally and now it doesn't work? 🤷‍♂

swalkinshaw:
Other than the one `environment` issue with `provision` everything seems good 👍

cmd/deploy.go
swalkinshaw: Fixed in master

cmd/droplet_create.go
TangRufus: Updated in https://github.com/roots/trellis-cli/pull/91/commits/0c2ec858bb93e1f2a3198cabb19ca9896fbcff8b

cmd/ad_hoc_playbook.go
```diff
@@ -0,0 +1,47 @@
...
+	"path/filepath"
+)
+
+type AdHocPlaybook struct {
+	Playbook
+	files map[string]string
+}
+
+func (p *AdHocPlaybook) Run(playbookYml string, args []string) error {
+	// TODO: Panic if files is empty.
```
swalkinshaw: Should this be done?

TangRufus: done in https://github.com/roots/trellis-cli/pull/91/commits/6000a33e558d8825a04e2d27c07cdb48258da8a9

PR #96, opened 2020-01-19
Fix deploy args

swalkinshaw:
These vars aren't being recognized by Ansible due to the quoting.

Looking at the existing deploy.sh script and they aren't quoted in there.

@TangRufus it makes sense for me to fix this first in master since it shouldn't be part of your playbook PR.

### DIFF
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -102,19 +102,19 @@ func TestDeployRun(t *testing.T) {
 		{
 			"default",
 			[]string{"development", "example.com"},
-			`ansible-playbook deploy.yml -e "env=development site=example.com"`,
+			"ansible-playbook deploy.yml -e env=development site=example.com",
 			0,
 		},
 		{
 			"site_not_needed_in_defaut_case",
 			[]string{"development"},
-			`ansible-playbook deploy.yml -e "env=development site=example.com"`,
+			"ansible-playbook deploy.yml -e env=development site=example.com",
 			0,
 		},
 		{
 			"with_extra_vars",
 			[]string{"-extra-vars", "k=v foo=bar", "development"},
-			`ansible-playbook deploy.yml -e "env=development site=example.com k=v foo=bar"`,
+			"ansible-playbook deploy.yml -e env=development site=example.com k=v foo=bar",
 			0,
 		},
 	}
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -73,7 +73,7 @@ func (c *DeployCommand) Run(args []string) int {
 		vars = append(vars, c.extraVars)
 	}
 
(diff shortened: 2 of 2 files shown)


cmd/deploy_test.go
```diff
@@ -102,19 +102,19 @@ func TestDeployRun(t *testing.T) {
 		{
 			"default",
 			[]string{"development", "example.com"},
-			`ansible-playbook deploy.yml -e "env=development site=example.com"`,
+			"ansible-playbook deploy.yml -e env=development site=example.com",
```
TangRufus: We need 1 `-e` per ansible variable, i.e: 

```
$ ansible-playbook deploy.yml -e env=development site=typist.tech
ERROR! the playbook: site=typist.tech could not be found

$ ansible-playbook deploy.yml -e env=development -e site=typist.tech
Work as expected
```

TangRufus: I believe the culprit is golang is not passing `-e` to `ansible-playbook`

Using trellis-cli 4c5e1eb

```sh-session
$ trellis deploy --extra-vars='a=b' production typist.tech
Running command => ansible-playbook deploy.yml -e "env=production site=typist.tech a=b"

PLAY [Ensure necessary variables are defined] **********************************

TASK [Ensure environment is defined] *******************************************
System info:
  Ansible 2.7.15; Darwin
  Trellis Head
---------------------------------------------------
Environment missing. Use `-e` to define `env`:
ansible-playbook deploy.yml -e "site=<domain> env=<environment>"

fatal: [localhost]: FAILED! => {"changed": false}
	to retry, use: --limit @/Users/rufus/Code/typist/trellis/deploy.retry

PLAY RECAP *********************************************************************
localhost                  : ok=0    changed=0    unreachable=0    failed=1
```

However, if I run the generated command directly:

```sh-session
$ ansible-playbook deploy.yml -e "env=production site=typist.tech a=b"
It works
```

Reason is unknown.

swalkinshaw: 😕 gotta dig into this more I guess

Thanks for testing

swalkinshaw: Wait I thought you were testing this branch? 4c5e1eb is master which has this known bug. Can you try with this branch? It's working for me

swalkinshaw:
Maybe we should ditch the annoying `--extra-args` and just pass through anything after as them?

ie: `trellis deploy production foo.com a=b`

Thought we support the no site option which makes it harder. We could detect `=` in the 2nd positional arg and treat it as extra args maybe. Or just keep the option flag 🤷‍♂

cmd/deploy_test.go
TangRufus: Oh............ I need coffee.
5a66436 works.
Sorry for wasting time

TangRufus:
I say just keep `--extra-args`

PR #97, opened 2020-01-19
Automatically install Galaxy roles with -f

swalkinshaw:
Closes #95 

This will automatically detect roles that are already installed but a different version than specified in the roles file (`galaxy.yml`) and install them for you.

This solves an annoying problem with `ansible-galaxy` because the easier  solution is to run `install --force` but that re-installs *everything* and is slower.

PR #98, opened 2020-01-19
Print the original stderr when first `ansible-galaxy install` fails

TangRufus:
https://github.com/roots/trellis-cli/pull/97#discussion_r368332716


swalkinshaw:
Good idea 👍

PR #100, opened 2020-01-21
Add `db open` command with `Playbook` struct

TangRufus:
Continue https://github.com/roots/trellis-cli/pull/72 (which accidentally mereged and then reverted on roots/trellis-cli#90)

Using `Playbook` struct from #91

See: #7


TangRufus:
All set. Please review.

PR #101, opened 2020-01-22
Readme: Update instruction for homebrew shell autocompletions

TangRufus:
Copied from https://github.com/heroku/homebrew-brew/blob/f4776ab5b43c531a2b1047a66cd45d1224031d46/Formula/heroku.rb#L18-L33

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,30 @@ trellis-cli provides binary releases for a variety of OSes. These binary version
 2. Unpack it (`tar -zxvf trellis_0.3.1_Linux_x86_64.tar.gz`)
 3. Find the `trellis` binary in the unpacked directory, and move it to its desired destination (`mv trellis_0.3.1_Darwin_x86_64/trellis /usr/local/bin/trellis`)
 4. Make sure the above path is in your `$PATH`
-5. Run `trellis --autocomplete-install` to install shell autocompletions
+
+### Shell Autocompletions
+
+Homebrew installs trellis-cli' shell completion automatically by default. If shell completions aren't working, or you installed manually not using Homebrew, you'll need to install the completions manually
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,7 +47,17 @@ trellis-cli provides binary releases for a variety of OSes. These binary version
 2. Unpack it (`tar -zxvf trellis_0.3.1_Linux_x86_64.tar.gz`)
 3. Find the `trellis` binary in the unpacked directory, and move it to its desired destination (`mv trellis_0.3.1_Darwin_x86_64/trellis /usr/local/bin/trellis`)
 4. Make sure the above path is in your `$PATH`
-5. Run `trellis --autocomplete-install` to install shell autocompletions
+
+### Shell Autocompletions
+
+To use the trellis-cli's autocomplete via Homebrew's shell completion:
```
swalkinshaw: Are these instructions meant for people who have installed trellis-cli via homebrew? If so, we already take care of this automatically: https://github.com/roots/homebrew-tap/blob/master/Formula/trellis-cli.rb#L23

If it's meant for non-homebrew installs the wording seems confusing to me.

TangRufus: Step 1 is the important bit.
Turns out I never set up Homebrew's shell completion properly.

Once step 1 is done, I can install trellis' completion by, either:
- `brew reinstall trellis-cli`, or
- `trellis --autocomplete-install`

I think give user more instructions doesn't hurt.
Re-running `trellis --autocomplete-install` doesn't hurt neither.

swalkinshaw: Maybe just mention that using Homebrew they *should* be installed automatically? Then something like

"If shell completions aren't working, or you installed manually not using Homebrew, you'll need to install the completions manually"

Issue #10, opened 2019-01-17
SSL migration tool


retlehs:
scenario: you have a trellis site with let's encrypt that's already live and you need to provision a new server

problem: let's encrypt cert issuing will fail since DNS isn't yet pointed to the new server

solution: temporarily copy the let's encrypt certs from the live site and use them on the new server until DNS is updated

retlehs:
the manual way:

```
scp admin@example.com:/etc/nginx/ssl/letsencrypt/example.com-bundled.cert /local/path
scp admin@example.com:/etc/nginx/ssl/letsencrypt/example.com.key /local/path
```

```
ssl:
  enabled: true
  # provider: letsencrypt
  provider: manual
  cert: /local/path/example.com-bundled.cert
  key: /local/path/example.com.key
```

strarsis:
@retlehs: Thanks for this guide! Though using the commands above to download the existing certs doesn't work straightforward as I get a permission denied error (`root` user needed), as `scp` doesn't support `sudo`.

retlehs:
@strarsis you're welcome, sorry it's just some rough notes haha! 

did you get the permission error when using the admin user?

strarsis:
@retlehs: Yes, I get this error also as admin user, because the `/etc/nginx/ssl/letsencrypt//` directory is read-only for `root`.

Twansparant:
> @retlehs: Yes, I get this error also as admin user, because the `/etc/nginx/ssl` directory is read-only for `root`.

Did you find a workaround for this? Switching to `root` user might fix it, but then you need enter the root's password?

strarsis:
@Twansparant: I used `sudo` (or `sudo su`), enterred the `admin` password (this classic no-root login, admin-sudo-user login setup is what Trellis uses) to get root permissions.

Twansparant:
Thanks @strarsis, I figured it out myself already but switching to root user prompted me for the root password so I had to reset that in DO. All good now, thanks for the help!

retlehs:
Ref https://discourse.roots.io/t/deploying-a-trellis-site-with-ssl-certificates-prior-to-dns-records-being-in-place/24201/9

PR #107, opened 2020-02-16
Add `SSLFetchCommand`

TangRufus:
Why not use ansible synchronize module?

Because it only supports passwordless sudo.

Why use ansible find module?

Because the fetch module doesn't support fetching directories recursively.

Related: https://github.com/roots/trellis-cli/issues/10

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ Supported commands so far:
 | `provision` | Provisions the specified environment |
 | `rollback` | Rollsback the last deploy of the site on the specified environment |
 | `ssh` | Connects to host via SSH |
+| `ssl` | Commands for SSL certificates management |
 | `up` | Starts and provisions the Vagrant environment by running `vagrant up` |
 | `valet` | Commands for Laravel Valet |
 | `vault` | Commands for Ansible Vault |
--- a/cmd/ssl_fetch.go
+++ b/cmd/ssl_fetch.go
@@ -1,0 +1,121 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/fatih/color"
+	"github.com/mitchellh/cli"
+	"strings"
+	"time"
+	"trellis-cli/trellis"
+)
+
+type SSLFetchCommand struct {
+	UI       cli.Ui
+	Trellis  *trellis.Trellis
+	playbook PlaybookRunner
+}
+
+const sslFetchYmlContent = `
+---
+- name: Test Connection and Determine Remote User
+  hosts: web:&{{ env }}
+  gather_facts: false
+  roles:
+    - { role: connection, tags: [connection, always] }
+
+- name: "Trellis CLI: Pull SSL certificates"
+  hosts: web:&{{ env }}
+  gather_facts: false
+  become: true
+  tasks:
+    - name: Conditionally 'nginx_path' variable from nginx role
+      include_vars:
+        file: roles/nginx/defaults/main.yml
+      when: nginx_path is not defined
+
+    - name: Fail if couldn't determine 'nginx_path' variable
+      fail:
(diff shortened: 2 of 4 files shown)


swalkinshaw:
Curious what your workflow is like for this use case?

TangRufus:
1. download the Let'sEncrypt certs from old servers
2. set Trellis SSL to manual
3. provision new servers
4. switch Trellis SSL back to Let'sEncrypt
5. reprovision

I must admit my experience with Let'sEncrypt on Trellis is limited.
I have only 1 Trellis site using Let'sEncrypt.

swalkinshaw:
Is there a step between 3 and 4 where you copy over the old certs you downloaded with this command?

TangRufus:
No.

To clarify (I reordered step 4 & 5 for clarity):

1. download the Let'sEncrypt certs from old servers <-- this PR
1. set Trellis SSL to manual (to use certs from step 1)
    ```yml
    ssl:
      enabled: true
      provider: manual
      cert: /path/to/my/cert/from/step/one
      key: /path/to/my/key/from/step/one
    ```
1. provision new servers (with certs from step 1 as `manual`)
1. change DNS records (new servers using old certs)
1. switch Trellis SSL back to Let'sEncrypt (no provision at this point; only re-provision **after** DNS propagation finsished)
1. reprovision new servers (when DNS propagation finsished)
1. destroy old servers


Note: This PR only does step 1.

My new servers could use the old certs without problems.

Am I missing something?

swalkinshaw:
Nope that makes sense and I forgot that feature of specifying the manual files 😅

I wonder if there's a slightly better step to transfer the old certs to the new one 🤔 the hard part is setting the provider to `manual` i think

TangRufus:
How about print a message (with a link to the docs) to suggest switching provider to `manual` at the end of the command?

swalkinshaw:
Adding a better after the command would be helpful. But also putting more instructions in the command description too 👍 

Here's what I'm thinking *might* be possible (without trying it or thinking about it too much 😅)

* building off of what's already here, the CLI downloads the certs
* then it *overrides* the new `ssl` config values for `manual` via Ansible's command line `-e` option which should take precedence over defined `group_vars`
* provisions the server with only a few tags necessary
* done?

TangRufus:
```sh-session
$ trellis ssl fetch
$ tree trellis-cli
trellis-cli
└── ssl
    └── fetch-20200216012015
        └── 111.222.333.444
            └── etc
                └── nginx
                    └── ssl
                        ├── client-4d77f9d.crt
                        └── cloudflare-origin-ca
                            ├── typist.tech.key
                            └── typist.tech.pem
```

I recommand having another command to do the migration and keep this one for those who just want to download the certs.

### Questions

### How do we get the paths to letsencrpty keys and certs?

https://github.com/roots/trellis/blob/f11a7cefbc978831d1ff6b4156298e618ff2ff80/roles/wordpress-setup/templates/wordpress-site.conf.j2#L95-L96

i.e: hjow to get `letsencrypt_cert_ids[item.key]` and/or [generate_cert_ids](https://github.com/roots/trellis/blob/52f7f6757314b92f123e2b418248b12626b777ad/roles/letsencrypt/tasks/certificates.yml#L21)

### Is downloading keys and certs from 1 of the web servers enough?

I remember Trellis' Let's Encrypt does special things when using multiple web servers (load balancers).

swalkinshaw:
> I recommand having another command to do the migration and keep this one for those who just want to download the certs.

Yeah makes sense 👍 

> How do we get the paths to letsencrpty keys and certs?

🤔 we could just get the `*.crt` and `*.key` files?

> Is downloading keys and certs from 1 of the web servers enough?

They should never differ so one should be correct

PR #110, opened 2020-03-03
Make autocompletion lazier

swalkinshaw:
These autocomplete functions are called *every* time the binary is run which was calling `LoadProject` in contexts we don't actually care if a project exists or not (like running `trellis help`, or any command which is project independent).

The only situations we actually want a project loaded is when we're autocompleting for the commands which is what will happen now since the checks are *within* the `PredictFunc`.

https://github.com/roots/trellis-cli/pull/65 was my previous "fix" for this but all that did was speed up successive **successful** look ups.

### DIFF
--- a/trellis/complete.go
+++ b/trellis/complete.go
@@ -5,23 +5,19 @@ import (
 )
 
 func (t *Trellis) AutocompleteSite() complete.Predictor {
-	if err := t.LoadProject(); err != nil {
-		return complete.PredictNothing
-	}
-
 	return t.PredictSite()
 }
 
 func (t *Trellis) AutocompleteEnvironment() complete.Predictor {
-	if err := t.LoadProject(); err != nil {
-		return complete.PredictNothing
-	}
-
 	return t.PredictEnvironment()
 }
 
 func (t *Trellis) PredictSite() complete.PredictFunc {
 	return func(args complete.Args) []string {
+		if err := t.LoadProject(); err != nil {
+			return []string{}
+		}
+
 		switch len(args.Completed) {
 		case 1:
 			return t.EnvironmentNames()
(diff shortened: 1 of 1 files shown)


TangRufus:
Code looks good to me.
However, I am unstable to test autocompletion on my machine.
When I press <kbd>tab</kbd>, zsh suggests files instead of environments or sites.

Shell completion works if I install via `$ brew install roots/tap/trellis-cli`
I am on macOS with zsh. Am I missing something?

Note: There is a great chance the culprit is I didn't set up my mac correctly. Ignore me if I am not making sense.

Tried:

```
$ git co fix-lazy-autocompletion
$ go build
$ ./trellis-cli --autocomplete-install
```

Also tried:

```
$ cat ~/Desktop/trellis-cli-dev.rb
class TrellisCliDev < Formula
  desc "A CLI to manage Trellis projects"
  homepage "https://roots.io/trellis"

  head "https://github.com/roots/trellis-cli.git", branch: "fix-lazy-autocompletion"
  depends_on "go"

  conflicts_with "roots/tap/trellis-cli"

  def install
    system "go", "build", "-o", "bin/trellis"
    bin.install "bin/trellis"
  end

  test do
    system "#{bin}/trellis --autocomplete-install"
    system "#{bin}/trellis -v"
  end
end


$ brew install --HEAD ~/Desktop/trellis-cli-dev.rb
```

swalkinshaw:
When I want to test it locally I run this in zsh:

```bash
complete -o nospace -C /Users/scott/dev/trellis-cli/trellis-cli trellis-cli
```

TangRufus:
I am being stupid again. `complete -o nospace -C xxxx` works!
The culprit is something in my `~/.zshrc` conflicts with `$ ./trellis-cli --autocomplete-install`

All working now! 🎉

PR #112, opened 2020-04-06
Run galaxy install on new project

swalkinshaw:
Also changes the `up` command to default to skipping galaxy install.
This changes the flag from `--no-galaxy` to `--with-galaxy`.

Since trellis-cli improves `galaxy install` and auto-runs on `provision`
and `new`, it makes sense to avoid the crappy `ansible-galaxy install`
run by Vagrant by default.


TangRufus:
Awesome 🎉

TangRufus:
Question: Do we want to add `galaxy install` on all playbook related commands, e.g: `trellis deploy`?

PR #114, opened 2020-04-17
Add shell hook for auto Virtualenv activation

swalkinshaw:
trellis-cli automatically creates a Virtualenv for managing Python and its dependencies and it's activated *within* trellis-cli for running commands. If you want to activate the Virtualenv *outside* of trellis-cli (in your shell), it either needs to be activated manually or every command needs to be run through `trellis exec`.

This adds automatic shell integration which will activate the Virtualenv transparently whenever you `cd` into a Trellis project and deactivates whenever you `cd` out of it.

A new `shell-init` command exists which prints out a script specific to your shell; Bash and Zsh are currently supported:

Zsh:
```bash
// Add to `.zshrc`
eval "$(trellis shell-init zsh)" 
```

Bash:
```bash
// Add to `.bash_profile`
eval "$(trellis shell-init bash)" 
```

### DIFF
--- a/cmd/shell_init.go
+++ b/cmd/shell_init.go
@@ -1,0 +1,209 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/mitchellh/cli"
+)
+
+const HookbookScript = `
+# Hookbook (https://github.com/Shopify/hookbook)
+#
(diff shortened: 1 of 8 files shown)


swalkinshaw:
@TangRufus would you be able to test this out please?

PR #115, opened 2020-04-21
fix(new): Fix double `v` on Trellis version

Log1x:
Trellis is versioned with `v` on GitHub where as Bedrock is not.

Screenshot

Trellis
Bedrock


swalkinshaw:
😱 good catch, thanks

PR #120, opened 2020-05-13
Re-add Windows releases

swalkinshaw:
We removed Windows support in https://github.com/roots/trellis-cli/pull/68 since our syscall usage wasn't compatible.

Now that `syscall.Exec` was removed in https://github.com/roots/trellis-cli/pull/118 we can build Windows binaries again.


swalkinshaw:
I'll have to update the README too

Issue #125, opened 2020-07-14
Feature request: Add `branch` option to deploy command


mike-sheppard:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing request
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Summary

Add a branch option to the `trellis deploy` command to override the deployed branch

Example:
```bash
trellis deloy staging example.com --branch=feature/branch-name
```

## Motivation

Simplify the deployment of custom branches. As a team we'll deploy our individual feature branches on staging before merging as part of our code review / QA process, it'd be great if we could simplify it with the `trellis-cli`.

## Additional context

Currently we achieve this by passing it through`-e EXTRA_VARS, --extra-vars` as discussed [in the discourse](https://discourse.roots.io/t/any-way-to-specify-branch-as-command-line-deploy-argument-to-override-default-branch-setting-in-wordpress-sites-yml/11937/2)

```diff
# group_vars/staging/wordpress_sites.yml
- branch: staging
+ branch: "{{ branch | default('staging') }}"
```

```bash
$ ansible-playbook deploy.yml -e "site=example.com env=staging branch=feature/new-stuff"
```

Thanks + this tool is v. much appreciated! :)

swalkinshaw:
I've opened a PR on Trellis as the first part of this: https://github.com/roots/trellis/pull/1204

PR #126, opened 2020-07-21
Add branch option to deploy

swalkinshaw:
Fixes https://github.com/roots/trellis-cli/issues/125

Not that this will only work if the Trellis project used has `branch` variable support from https://github.com/roots/trellis/pull/1204. There's no real way to set a dependency on a trellis-cli feature for a Trellis version yet.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -32,7 +32,7 @@ changelog:
     - '^test:'
 brews:
   - name: trellis-cli
-    github:
+    tap:
       owner: roots
       name: homebrew-tap
     folder: Formula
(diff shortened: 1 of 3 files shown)


.goreleaser.yml
```diff
@@ -32,7 +32,7 @@ changelog:
     - '^test:'
 brews:
   - name: trellis-cli
-    github:
+    tap:
```
TangRufus, line 35: Do we need to update goreleaser? Or, is it broken for a long time?

swalkinshaw: They often update/deprecate config options so I'm guessing they changed this after the last release we did. As far as I know every release has worked fine (but won't from now on until this fix).

swalkinshaw:
Merging this since `branch` variable is supported in Trellis 1.5.0 now.

PR #127, opened 2020-07-27
Update command name in README.md

runofthemill:
Just started checking out the CLI - seems awesome so far!

After adding `eval "$(trellis-cli shell-init zsh)"` to my `.zshrc` file:

```
❯ source ~/.zshrc
/Users/jeremy/.zshrc:117: command not found: trellis-cli
```

After adjusting to `eval "$(trellis shell-init zsh)"`:

```
❯ source ~/.zshrc

[trellis] activated env
```

Great success!

I also adjusted the header levels to hopefully make the hierarchy a little clearer; at first I thought the virtualenv integration was under the manual install header, so I wouldn't need to do it myself (having installed from Homebrew). This part's not as important, so I can also take out of the PR :)


swalkinshaw:
Good catch 🤦 thanks! Header fix seems better too

PR #136, opened 2020-10-30
Add version update notices

swalkinshaw:
This checks for newer versions of trellis-cli on every run (when enabled) and displays a notice when there's on available.

This is borrowed almost exactly from [GitHub's CLI](https://github.com/cli/cli). It's optimized to only check once every 24 hours and the result is cached locally to avoid HTTP fetches.

image

~I haven't added any tests yet... because it's fairly hard to test. But I'll see what I can do.~

Refactored this and now its testable! Amazing what good software design lets you do 🙃


config/config.go
```diff
@@ -0,0 +1,7 @@
+package config
+
+import (
+	gap "github.com/muesli/go-app-paths"
+)
+
+var Scope *gap.Scope = gap.NewScope(gap.User, "trellis")
```
swalkinshaw, line 7: I'm going to extract this into its own PR. trellis-cli sort of uses the XDG spec for storing the virtualenv but it's not correct. So now any config/data/cache file usage should go through this package which implements the XDG spec properly.

main.go
```diff
@@ -1,16 +1,33 @@
...
+	"trellis-cli/github"
 	"trellis-cli/trellis"
+	"trellis-cli/update"
 
+	"github.com/fatih/color"
+	"github.com/mattn/go-isatty"
 	"github.com/mitchellh/cli"
 )
 
+const version = "0.8.1"
```
swalkinshaw, line 19: For testing, will fix later

config/config.go
swalkinshaw: Eh I'm lazy. I'll update the other usage after this

main.go
```diff
@@ -1,16 +1,42 @@
...
 	"trellis-cli/trellis"
+	"trellis-cli/update"
 
+	"github.com/fatih/color"
 	"github.com/mitchellh/cli"
 )
 
+const version = "0.9.0"
+
+var updaterEnabled = ""
```
TangRufus, line 21:
```suggestion
var updaterEnabled = "roots/trellis-cli"
```

Or, was it left blank on purpose?

update/main.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	return true
+}
+
+func (n *Notifier) getLatestReleaseInfo(stateFilePath string) (*github.Release, error) {
+	stateEntry, err := getStateEntry(stateFilePath)
+	if err == nil && time.Since(stateEntry.CheckedForUpdateAt).Hours() < 24 {
+		return stateEntry.LatestRelease, nil
+	}
```
TangRufus, line 82: Should we force to fetch from GitHub if `TRELLIS_FORCE_UPDATE_NOTIFIER` is set?

TangRufus (changes requested):

main.go
swalkinshaw: Left blank on purpose so that normal builds (just using `go build`) don't include HTTP calls for the update checks.

Oh!! that reminds me I need to update `.gorelease.yml` to use `go build -ldflags="-X main.updaterEnabled=roots/trellis-cli"`

But yeah... I took the name from GitHub's and I agree it should have `repo` in it.

update/main.go
swalkinshaw: 🤔 it's a decent idea but it breaks my usage of that flag unfortunately. It mainly exists for testing purposes to get past the "is TTY?" check.

TangRufus: Maybe add a `force` param on `CheckForUpdate`?

TangRufus (approved):

update/main.go
swalkinshaw: Your idea was better than the env flag though as a direct replacement. So I refactored `TRELLIS_FORCE_UPDATE_NOTIFIER` to instead be a `ForceCheck` attribute on the notifier struct. And when I was redoing the tests, it exposed an issue 🎉  All fixed up now

PR #139, opened 2020-11-10
`InitCommand`: Show `pip` stack trace when failed

TangRufus:
Goal: To provide more helpful error messages

Before:

```sh-session
$ trellis init
Creating virtualenv in /Users/rufus/Code/typist/trellis/.trellis/virtualenv
✓ Virtualenv created

✓ Dependencies installed
Error installing pip requirements: exit status 1
```

After:

```sh-session
$ trellis init
Creating virtualenv in /Users/rufus/Code/typist/trellis/.trellis/virtualenv
✓ Virtualenv created

✓ Dependencies installed
Error installing pip requirements: exit status 1
Stack trace:
Requirement already satisfied: ansible<3.0,>=2.8.0 in ./.trellis/virtualenv/lib/python3.9/site-packages (from -r requirements.txt (line 1)) (2.10.1)
Requirement already satisfied: passlib in ./.trellis/virtualenv/lib/python3.9/site-packages (from -r requirements.txt (line 2)) (1.7.4)
ERROR: Could not find a version that satisfies the requirement fake-package-to-trigger-error (from -r requirements.txt (line 3)) (from versions: none)
ERROR: No matching distribution found for fake-package-to-trigger-error (from -r requirements.txt (line 3))
WARNING: You are using pip version 20.2.3; however, version 20.2.4 is available.
You should consider upgrading via the '/Users/rufus/Code/typist/trellis/.trellis/virtualenv/bin/python3 -m pip install --upgrade pip' command.
```

To test: Add a fake package in [`requirements.txt`](https://github.com/roots/trellis/blob/2ae8f383af42da9d37bce91531c4175c9d1b6d9f/requirements.txt)

To discuss: 
- Should we do the same on other `exec.Command.Run()`?
- Alternatively, we can use `exec.Command.Output()`
- Is that a way to _correct_ console output coloring like `$ trellis exec pip install -r requirements.txt`?

cc @mariusz-ah

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -2,7 +2,7 @@ package cmd
 
 import (
 	"fmt"
-	"os/exec"
+	"github.com/fatih/color"
 	"strings"
 	"time"
 
@@ -39,10 +39,10 @@ func (c *InitCommand) Run(args []string) int {
 		c.UI.Info("virtualenv not found")
 		s := spinner.New(spinner.CharSets[14], 100*time.Millisecond)
 		s.Suffix = " Installing virtualenv..."
-		s.FinalMSG = "✓ virtualenv installed\n"
 		s.Start()
 		c.Trellis.Virtualenv.Install()
 		s.Stop()
+		c.UI.Info(color.GreenString("✓ virtualenv installed"))
 	}
 
 	c.UI.Info(fmt.Sprintf("Creating virtualenv in %s", c.Trellis.Virtualenv.Path))
@@ -53,24 +53,22 @@ func (c *InitCommand) Run(args []string) int {
 		return 1
 	}
 
-	c.UI.Info("✓ Virtualenv created\n")
+	c.UI.Info(color.GreenString("✓ Virtualenv created"))
 
-	pip := exec.Command("pip", "install", "-r", "requirements.txt")
+	pip := CommandExecWithStderrOnly("pip", []string{"install", "-r", "requirements.txt"}, c.UI)
 
 	s := spinner.New(spinner.CharSets[14], 100*time.Millisecond)
 	s.Suffix = " Installing pip dependencies (pip install -r requirements.txt) ..."
-	s.FinalMSG = "✓ Dependencies installed\n"
 	s.Start()
 
 	err = pip.Run()
+	s.Stop()
 
 	if err != nil {
-		s.Stop()
-		c.UI.Error(fmt.Sprintf("Error installing pip requirements: %s", err))
 		return 1
 	}
 
-	s.Stop()
+	c.UI.Info(color.GreenString("✓ Dependencies installed"))
 
 	return 0
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -37,6 +37,15 @@ func CommandExecWithOutput(command string, args []string, ui cli.Ui) *exec.Cmd {
 	return cmd
 }
 
+func CommandExecWithStderrOnly(command string, args []string, ui cli.Ui) *exec.Cmd {
+	cmd := exec.Command(command, args...)
+	cmd.Stderr = os.Stderr
+
+	ui.Info(fmt.Sprintf("Running command => %s", strings.Join(cmd.Args, " ")))
+
+	return cmd
+}
+
 func CommandExec(command string, args []string, ui cli.Ui) *exec.Cmd {
 	cmd := exec.Command(command, args...)
 	cmd.Stdin = os.Stdin


cmd/init.go
```diff
@@ -62,11 +62,12 @@ func (c *InitCommand) Run(args []string) int {
 	s.FinalMSG = "✓ Dependencies installed\n"
```
swalkinshaw, line 62: 🤦 I just noticed from your screenshot this outputs regardless of success or failure which isn't good. There might be an easy fix for that 🤔

cmd/init.go
```diff
@@ -62,11 +62,12 @@ func (c *InitCommand) Run(args []string) int {
...
 	s.Start()
 
-	err = pip.Run()
+	pipCombinedOutput, err := pip.CombinedOutput()
 
 	if err != nil {
 		s.Stop()
 		c.UI.Error(fmt.Sprintf("Error installing pip requirements: %s", err))
```
swalkinshaw, line 69: I don't think this line is really useful (or ever was) since it just shows the exist status. Let's just combine these?

```go
c.UI.Error(fmt.Sprintf("Error installing pip requirements: \n%s", pipCombinedOutput))
```

swalkinshaw:
Thanks for bringing this up. The experience is pretty bad right now!

> Should we do the same on other `exec.Command.Run()`?
In some places already I think it explicitly returns error output? Here's an example: https://github.com/roots/trellis-cli/blob/f3df24ddc3ec3147b9c6d232af5fae694a3179d9/cmd/galaxy_install.go#L53-L62

Or anywhere that uses `execCommandWithOutput`

> Alternatively, we can use `exec.Command.Output()`
I think `execCommandWithOutput` is that?

> Is that a way to _correct_ console output coloring like `$ trellis exec pip install -r requirements.txt`?
What do you mean by "correct"? If you run that commend above, is there just no coloring?

cmd/init.go
swalkinshaw: I think this might work:

```diff
        pip := exec.Command("pip", "install", "-r", "requirements.txt")

+       var finalMsg string
        s := spinner.New(spinner.CharSets[14], 100*time.Millisecond)
        s.Suffix = " Installing pip dependencies (pip install -r requirements.txt) ..."
-       s.FinalMSG = "✓ Dependencies installed\n"
+       s.FinalMSG = finalMsg
        s.Start()

-       err = pip.Run()
+       pipCombinedOutput, err := pip.CombinedOutput()

        if err != nil {
+               finalMsg = "✗ Dependencies not installed\n"
                s.Stop()
-               c.UI.Error(fmt.Sprintf("Error installing pip requirements: %s", err))
+               c.UI.Error(fmt.Sprintf("Error installing pip requirements: \n%s", pipCombinedOutput))
                return 1
        }

+       finalMsg = "✓ Dependencies installed\n"
        s.Stop()

        return 0
```

TangRufus:
Updated to use `stderr` directly and colorized the messages:

new-pr

cmd/init.go
```diff
@@ -39,10 +39,10 @@ func (c *InitCommand) Run(args []string) int {
 		c.UI.Info("virtualenv not found")
 		s := spinner.New(spinner.CharSets[14], 100*time.Millisecond)
 		s.Suffix = " Installing virtualenv..."
-		s.FinalMSG = "✓ virtualenv installed\n"
 		s.Start()
 		c.Trellis.Virtualenv.Install()
 		s.Stop()
+		c.UI.Info(color.GreenString("✓ virtualenv installed"))
```
swalkinshaw, line 45: Oh interesting... I guess `FinalMSG` is kind of useless in this case so we can just log as normal after 👍

cmd/init.go
```diff
@@ -53,24 +53,22 @@ func (c *InitCommand) Run(args []string) int {
 		return 1
 	}
 
-	c.UI.Info("✓ Virtualenv created\n")
+	c.UI.Info(color.GreenString("✓ Virtualenv created"))
 
-	pip := exec.Command("pip", "install", "-r", "requirements.txt")
+	pip := CommandExecWithStderrOnly("pip", []string{"install", "-r", "requirements.txt"}, c.UI)
```
swalkinshaw, line 58: What about copying how `galaxy install` works?

https://github.com/roots/trellis-cli/blob/f17dd81ec3320559b56bdb98b05f7b7b9f3aff8a/cmd/galaxy_install.go#L53-L64

Then we don't need the new `CommandExecWithStderrOnly` and you have more control over the output to add the newlines?

TangRufus: The goal of `CommandExecWithStderrOnly` is to retain coloring of the wanring messages. Here is a screenshot of `CommandExecWithStderrOnly` vs the `galaxy install` way:

trellis-init-coloring

swalkinshaw: Weird, I tried it out and got the same. Found this https://stackoverflow.com/questions/61113201/io-multiwriter-removes-terminal-colors-from-output but it wasn't really any help.

Okay, interesting I think this is why https://stackoverflow.com/a/29375330

`pip` only has a `--no-color` option but not way to force it. Who knows if there's actually a solution so what you have works 👍

swalkinshaw (approved):

PR #143, opened 2020-11-23
Fix #142 - Quote/escape paths in venv hook

### LINKED ISSUE #142: Trellis virtualenv/shell integration breaks with spaces in $PATH
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)

## Description

With a `$PATH` that contains a space, the `virtualenv` integration with `bash` trips on updating `$PATH` correctly:

```
$ eval "$(trellis shell-init bash)"
$ set -xv
$ cd abundanthousingma.org/
cd abundanthousingma.org/
__hookbook_debug_trap "$_"
++ __hookbook_debug_trap -xv
+ cd abundanthousingma.org/
{
        [[ $- =~ x ]] && {
          \set +x; __trellis_cli_hook precmd 2>&3; \set -x;
        } || {
          __trellis_cli_hook precmd 2>&3;
        }
      } 4>&2 2>/dev/null 3>&4;
[trellis] activated env
export VIRTUAL_ENV=/Users/jeffbyrnes/dev/abundanthousingma.org/trellis/.trellis/virtualenv
export PRE_TRELLIS_PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/Little Snitch.app/Contents/Components:/Library/Apple/usr/bin
-bash: export: `Snitch.app/Contents/Components:/Library/Apple/usr/bin': not a valid identifier
export PATH=/Users/jeffbyrnes/dev/abundanthousingma.org/trellis/.trellis/virtualenv/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/Little Snitch.app/Contents/Components:/Library/Apple/usr/bin
-bash: export: `Snitch.app/Contents/Components:/Library/Apple/usr/bin': not a valid identifier
```

## Steps to reproduce

1. Add a path to `$PATH` that contains a space
2. Enable the Trellis `virtualenv` integration
3. Switch to the project root

**Expected behavior:** The `virtualenv` integration works well.

**Actual behavior:** The integration incorrectly sets `$PATH`

**Reproduces how often:** Always

## Versions

```
$ trellis -v
0.9.1
```

swalkinshaw:
Without quoting (escaping) paths with spaces (or potentially other special characters) would break the hook script.

cc @jeffbyrnes

### DIFF
--- a/cmd/venv_hook.go
+++ b/cmd/venv_hook.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/mitchellh/cli"
+	"gopkg.in/alessio/shellescape.v1"
 	"trellis-cli/trellis"
 )
 
@@ -21,9 +22,9 @@ func (c *VenvHookCommand) Run(args []string) int {
 	if c.Trellis.ActivateProject() {
 		if !ok {
 			fmt.Fprintf(color.Error, "[trellis] \x1b[1;32mactivated env\x1b[0m\n")
-			c.UI.Output(fmt.Sprintf("export %s=%s", trellis.VenvEnvName, c.Trellis.Virtualenv.Path))
-			c.UI.Output(fmt.Sprintf("export %s=%s", trellis.OldPathEnvName, c.Trellis.Virtualenv.OldPath))
-			c.UI.Output(fmt.Sprintf("export %s=%s:%s", trellis.PathEnvName, c.Trellis.Virtualenv.BinPath, c.Trellis.Virtualenv.OldPath))
+			c.UI.Output(fmt.Sprintf("export %s=%s", trellis.VenvEnvName, shellescape.Quote(c.Trellis.Virtualenv.Path)))
+			c.UI.Output(fmt.Sprintf("export %s=%s", trellis.OldPathEnvName, shellescape.Quote(c.Trellis.Virtualenv.OldPath)))
+			c.UI.Output(fmt.Sprintf("export %s=%s:%s", trellis.PathEnvName, c.Trellis.Virtualenv.BinPath, shellescape.Quote(c.Trellis.Virtualenv.OldPath)))
 		}
 	} else {
 		if ok {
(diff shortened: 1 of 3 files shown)


cmd/venv_hook.go
```diff
@@ -21,17 +22,17 @@ func (c *VenvHookCommand) Run(args []string) int {
...
+			c.UI.Output(fmt.Sprintf("export %s=%s", trellis.VenvEnvName, shellescape.Quote(c.Trellis.Virtualenv.Path)))
+			c.UI.Output(fmt.Sprintf("export %s=%s", trellis.OldPathEnvName, shellescape.Quote(c.Trellis.Virtualenv.OldPath)))
+			c.UI.Output(fmt.Sprintf("export %s=%s:%s", trellis.PathEnvName, c.Trellis.Virtualenv.BinPath, shellescape.Quote(c.Trellis.Virtualenv.OldPath)))
 		}
 	} else {
 		if ok {
 			fmt.Fprintf(color.Error, "[trellis] \x1b[1;31mdeactivated env\x1b[0m\n")
 			path := os.Getenv(trellis.OldPathEnvName)
 			c.UI.Output(fmt.Sprintf("unset %s", trellis.VenvEnvName))
 			c.UI.Output(fmt.Sprintf("unset %s", trellis.OldPathEnvName))
```
TangRufus, line 34: Do we need escape here as well?

Or, is `unset` a command or just logging?

swalkinshaw: It's a command, but `unset` only takes the key/name and not a value as well so no need to escape.

TangRufus (approved):

TangRufus:
Side questions: 
- When to use `os.Setenv`? When to use `c.UI.Output(fmt.Sprintf("export %s=%s", ......`
- When to use `os.Unsetenv`? When to use `c.UI.Output(fmt.Sprintf("unset %s", ......`

swalkinshaw:
The virtualenv shell integration is fairly complex so it's confusing (I had to remind myself to fix this bug), but here's the overview:

* `shell-init` prints a shell script which is the entrypoint for the shell that hooks into `cd` (basically)
* currently the above script just runs `trellis venv hook` (it could run other kinds of shell integration in the future)
* `venv hook` also outputs a shell script

So all of that is solely for shell integration which is why it needs to output a shell script/commands hence the `c.UI.Output(fmt.Sprintf("export %s=%s", ......`. These get evaluated in the shell, not in Go.

Controlling env vars within Go (via `Setenv`, `Unsetenv`, etc) are used for the virtualenv integration *within* the `trellis` program itself. So trellis-cli internally activates the virtualenv for itself, and that has to use the Go env commands.

Hope that makes sense.

TangRufus:
Looks good.
Let discuss when we have more use cases of virtualenv shell integration.

jeffbyrnes:
Fantastic! This definitely looks like it’ll solve things, and it makes sense that this wasn’t working, since it outputs it to the shell for evaluation.

swalkinshaw:
I'll do a release later tonight

PR #144, opened 2020-11-30
[Help Wanted] Add plugin system (Look for `trellis-*` in `$PATH`)

### LINKED ISSUE #36: Plugins/3rd party commands
This topic has come up a few times and I've found some solutions recently. They call into two categories:

* native Go `plugin`
* RPC

## Go Plugins

Go actually has a `Plugin` package for shared libraries which makes this pretty easy: https://golang.org/pkg/plugin/

Pros: built-in, fast, simpler?
Cons: ~it kind of exists on Windows~ Linux/macOS only
## RPC

This is how a lot of CLIs solve the problem. They do RPC communication between processes. https://github.com/hashicorp/go-plugin looks great for this.

Pros: easier to distribute (normal build process), all platforms
Cons: slower, bit more complicated to build

Speed/performance doesn't really matter much for our purposes.

Right now I'm leaning towards the `plugin` package and punting on Windows support for 3rd party plugins.

cc @tangrufus

TangRufus:
Inspired by https://github.com/kubernetes/kubectl/blob/b155278f1f4a21a0be2d4f6f0037258dee4d1a22/pkg/cmd/plugin/plugin.go and https://github.com/kubernetes/kubectl/blob/b155278f1f4a21a0be2d4f6f0037258dee4d1a22/pkg/cmd/cmd.go#L371

To test:
```bash
cd trellis-cli
go build

mkdir testing
ln -s $(which git) ./testing/trellis-git
echo "#\!/bin/bash\necho 'i am plugintest'" > testing/trellis-plugintest
echo "#\!/bin/bash\necho 'i am theplugin subtest'" > testing/trellis-theplugin-subtest
sudo chmod +x testing/trellis-plugintest
sudo chmod +x testing/trellis-theplugin-subtest


PATH="$(pwd)/testing:$PATH" ./trellis-cli
PATH="$(pwd)/testing:$PATH" ./trellis-cli plugintest
PATH="$(pwd)/testing:$PATH" ./trellis-cli theplugin subtest
PATH="$(pwd)/testing:$PATH" ./trellis-cli git 
PATH="$(pwd)/testing:$PATH" ./trellis-cli git log
PATH="$(pwd)/testing:$PATH" ./trellis-cli git log --help
PATH="$(pwd)/testing:$PATH" ./trellis-cli git help log
```

Related: #37 #145 
Close #36

### DIFF
--- a/cmd/passthrough.go
+++ b/cmd/passthrough.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2014 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"syscall"
+)
+
+type PassthroughCommand struct {
+	Bin  string
+	Name string
+	Args []string
+}
+
+// Taken from https://github.com/kubernetes/kubectl/blob/b155278f1f4a21a0be2d4f6f0037258dee4d1a22/pkg/cmd/cmd.go#L371
+func (c *PassthroughCommand) Run(args []string) int {
+	// Windows does not support exec syscall.
+	if runtime.GOOS == "windows" {
+		cmd := exec.Command(c.Bin, args...)
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		cmd.Stdin = os.Stdin
+		cmd.Env = os.Environ()
+		err := cmd.Run()
+		if err == nil {
+			return 0
+		}
+		return 1
+	}
+
+	// invoke cmd binary relaying the environment and args given
+	// append executablePath to cmdArgs, as execve will make first argument the "binary name".
+	if err := syscall.Exec(c.Bin, append([]string{c.Bin}, args...), os.Environ()); err != nil {
+		return 1
+	}
+
+	return 0
+}
+
+func (c *PassthroughCommand) Synopsis() string {
+	return fmt.Sprintf("Third party plugin: Forward command to %s", filepath.Base(c.Bin))
+}
+
+func (c *PassthroughCommand) Help() string {
+	requested := strings.Join(c.Args, " ")
+
+	if strings.HasPrefix(requested, c.Name) {
+		commandParts := len(strings.Split(c.Name, " "))
+		c.Run(c.Args[commandParts:])
+	}
+
+	return ""
+}
--- a/plugin/finder.go
+++ b/plugin/finder.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package plugin
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+type finder struct {
+	validPrefixes    []string
+	searchPaths      []string
+	coreRootCommands []string
+}
+
+func (o *finder) find() map[string]string {
+	plugins := make(map[string]string)
+
+	for _, dir := range unique(o.searchPaths) {
+		if len(strings.TrimSpace(dir)) == 0 {
+			continue
+		}
+
+		files, err := ioutil.ReadDir(dir)
+		if err != nil {
+			continue
+		}
+
+		for _, f := range files {
+			if f.IsDir() {
+				continue
+			}
+			if !hasValidPrefix(f.Name(), o.validPrefixes) {
+				continue
+			}
+
+			// Prevent overriding core commands or adding any subcommands under core commands.
+			if isUnderCoreRootCommands(f.Name(), o.coreRootCommands) {
+				continue
+			}
+
+			path := filepath.Join(dir, f.Name())
+			if !isExecutable(path) {
+				continue
+			}
+
+			nameParts := strings.Split(f.Name(), "-")
+			if len(nameParts) > 1 {
+				// the first argument is always one of `validPrefixes` (i.e: "trellis") for a plugin binary
+				nameParts = nameParts[1:]
+			}
+			name := strings.Join(nameParts, " ")
+
+			// TODO: Handle overshadowed plugins, i.e: plugins with the same binary name.
+			plugins[name] = path
+		}
+	}
+
+	return plugins
+}
+
+func isExecutable(fullPath string) bool {
+	info, err := os.Stat(fullPath)
+	if err != nil {
+		return false
+	}
+
+	if runtime.GOOS == "windows" {
+		fileExt := strings.ToLower(filepath.Ext(fullPath))
+
+		switch fileExt {
+		case ".bat", ".cmd", ".com", ".exe", ".ps1":
+			return true
+		}
+		return false
+	}
+
+	if m := info.Mode(); !m.IsDir() && m&0111 != 0 {
+		return true
+	}
+
+	return false
+}
+
+func hasValidPrefix(filepath string, validPrefixes []string) bool {
+	for _, prefix := range validPrefixes {
+		if !strings.HasPrefix(filepath, prefix+"-") {
+			continue
+		}
+		return true
+	}
+	return false
+}
+
+func isUnderCoreRootCommands(filepath string, coreRootCommands []string) bool {
+	// the first argument is always one of `validPrefixes` (i.e: "trellis") for a plugin binary
+	rootCommand := strings.Split(filepath, "-")[1]
+
+	for _, v := range coreRootCommands {
+		if v == rootCommand {
+			return true
+		}
+	}
+	return false
+}
--- a/plugin/help_func.go
+++ b/plugin/help_func.go
@@ -1,0 +1,22 @@
+package plugin
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/mitchellh/cli"
+)
+
+func helpFunc(app string, pluginRootCommands []string) cli.HelpFunc {
+	return func(commands map[string]cli.CommandFactory) string {
+		var buf bytes.Buffer
+		if len(pluginRootCommands) > 0 {
+			buf.WriteString("\n\nAvailable third party plugin commands are:\n")
+		}
+
+		for _, p := range pluginRootCommands {
+			buf.WriteString(fmt.Sprintf("    %s\n", p))
+		}
+
+		return cli.BasicHelpFunc(app)(commands) + buf.String()
+	}
+}
--- a/plugin/register.go
+++ b/plugin/register.go
@@ -1,0 +1,67 @@
+package plugin
+
+import (
+	"github.com/mitchellh/cli"
+	"reflect"
+	"strings"
+	"trellis-cli/cmd"
+)
+
+func Register(c *cli.CLI, searchPaths []string, validPluginFilenamePrefixes []string) {
+	coreRootCommands := rootCommandsFor(reflect.ValueOf(c.Commands))
+
+	pluginFinder := finder{
+		validPrefixes:    validPluginFilenamePrefixes,
+		searchPaths:      searchPaths,
+		coreRootCommands: coreRootCommands,
+	}
+	plugins := pluginFinder.find()
+	pluginRootCommands := rootCommandsFor(reflect.ValueOf(plugins))
+
+	// Register plugin commands.
+	for name, bin := range plugins {
+		pluginCommand := &cmd.PassthroughCommand{
+			Name: name,
+			Bin:  bin,
+			Args: c.Args,
+		}
+		c.Commands[name] = func() (cli.Command, error) {
+			return pluginCommand, nil
+		}
+	}
+
+	// Separate plugin commands from core command lists.
+	c.HiddenCommands = append(c.HiddenCommands, pluginRootCommands...)
+	// Append plugin command list to help text.
+	c.HelpFunc = helpFunc(c.Name, pluginRootCommands)
+}
+
+func rootCommandsFor(v reflect.Value) (rootCommands []string) {
+	m := v.MapKeys()
+
+	for _, v := range m {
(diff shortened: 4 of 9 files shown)


TangRufus:
#### Help Wanted

- Write tests
- Give `plugin/namespaces.go` better names
- Test on linux
- Test on windows
- Skip the whole plugin system when `TRELLIS_NO_PLUGINS` environment variable is true
- Code review / refactor

#### Discussion 

Should we allow plugin to override core commands?
For example: if `trellis-check` exisits in `$PATH`, it will override the default one.

plugin/namespaces.go
```diff
@@ -0,0 +1,84 @@
...
+		for i, _ := range make([]int, len(segments)) {
+			namespace := strings.Join(segments[:i+1], " ")
+			namespaces = append(namespaces, namespace)
+		}
+	}
+
+	return missing(n.displays, namespaces)
+}
+
+// TODO: Question: Any way to accept both `map[string]string` and `map[string]cli.CommandFactory` at the same time?
```
swalkinshaw, line 39: `map[string]interface{}` is the only way to do afaik. https://bitfieldconsulting.com/golang/map-string-interface

plugin/namespaces.go
```diff
@@ -0,0 +1,84 @@
...
+func keysForStringMap(m map[string]string) []string {
+	keys := make([]string, len(m))
+
+	i := 0
+	for k := range m {
```
swalkinshaw, line 44:
```go
for i, k:= range m {
  keys[i] = k
}
```

Should handle the index without doing it manually

plugin/namespaces.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return keys
+}
+
+func keysForCommandFactoryMap(m map[string]cli.CommandFactory) []string {
+	keys := make([]string, len(m))
+
+	i := 0
+	for k := range m {
```
swalkinshaw, line 56: ditto here

plugin/help_func.go
```diff
@@ -0,0 +1,22 @@
...
+
+func helpFunc(app string, plugins map[string]string) cli.HelpFunc {
+	return func(commands map[string]cli.CommandFactory) string {
+		var buf bytes.Buffer
+		if len(plugins) > 0 {
+			buf.WriteString("\n\nAvailable plugins commands are:\n")
+		}
+
+		for name, _ := range plugins {
+			buf.WriteString(fmt.Sprintf("    %s\n", name))
```
swalkinshaw, line 17: This just shows the name right? I'm guessing this is missing actual help output from the plugin commands so far?

TangRufus: Just the names.
If we want the help text, we will require plugins to have either:
a. a man page we could parse, or
b. a special command to return short synopsis, e.g: `$ trellis-my-plugin --synopsis`, or
c. some way to register synopsis like #145 

However, I think this is an overkill at the monment. How about we revisit this when we have thrid party plugins?

plugin/namespaces.go
TangRufus: Because we want to turn the map into a slice, and slice keys must be intergers.

In `for i, k:= range m {`, `i` is the string key we want to collect, `k` is the vaule we dont need.

See: https://play.golang.org/p/_-rKpEraZt6

plugin/namespaces.go
TangRufus: `[]string` and `CommandFactory` cannot be cast to `interface{}`:

```sh-session
$ go build
# trellis-cli/plugin
plugin/namespaces.go:15:35: cannot use coreCommands (type map[string]cli.CommandFactory) as type map[string]interface {} in argument to keysForCommandFactoryMap
plugin/namespaces.go:17:29: cannot use pluginCommands (type map[string]string) as type map[string]interface {} in argument to keysForStringMap
```

Using `reflect.ValueOf(xxx).MapKeys()` in [`d600066` (#144)](https://github.com/roots/trellis-cli/pull/144/commits/d60006605098bf7a06dee6f8d25abc8c022b64a9)

TangRufus:
Note: Unlike the discarded commits in #136, we use [`strconv.ParseBool`](https://golang.org/pkg/strconv/#ParseBool) to parse `TRELLIS_NO_PLUGINS`.
It accepts `1, t, T, TRUE, true, True` only

plugin/namespaces.go
swalkinshaw: Ah sorry, I might do this then:

```go
keys := []string{}

for k := range m {
  keys = append(keys, k)
}

return keys
```

It reads better to me and is a little simpler?

cmd/passthrough.go
```diff
@@ -0,0 +1,78 @@
...
+		err := cmd.Run()
+		if err == nil {
+			return 0
+		}
+		return 1
+	}
+
+	// invoke cmd binary relaying the environment and args given
+	// append executablePath to cmdArgs, as execve will make first argument the "binary name".
+	if err := syscall.Exec(c.bin, append([]string{c.bin}, args...), os.Environ()); err != nil {
```
swalkinshaw, line 58: Do you know if this method is preferred to the windows "workaround" above?

If the windows method works fine, couldn't we just use that for all platforms unconditionally?

cmd/passthrough.go
```diff
@@ -0,0 +1,78 @@
...
+		return 1
+	}
+
+	return 0
+}
+
+func (c *PassthroughCommand) Synopsis() string {
+	return fmt.Sprintf("Thrid party plugin: Forawrd command to %s", filepath.Base(c.bin))
```
swalkinshaw, line 66: "Third" typo and "Forward"

plugin/finder.go
```diff
@@ -0,0 +1,144 @@
...
+
+type binPathVerifier struct{}
+
+// verify implements pathVerifier and determines if a given path
+// is valid depending on whether or not it overwrites an existing
+// trellis command path, or a previously seen plugin.
+func (v *binPathVerifier) verify(path string) (bool, string, string) {
+	// extract the plugin binary name
+	segments := strings.Split(path, "/")
+	binName := segments[len(segments)-1]
```
swalkinshaw, line 83: `filepath.Base(path)` instead? It handles Windows path separators too.

plugin/help_func.go
swalkinshaw: Yeah if that's the only downside then I think it's fine to skip for now as well 👍

swalkinshaw:
> Should we allow plugin to override core commands?

I think I'd prefer to skip this for now. I don't want people getting confused *which* subcommand is actually being run and not knowing if it's some custom override or not.

plugin/namespaces.go
```diff
@@ -0,0 +1,71 @@
+package plugin
+
+import (
+	"fmt"
+	"github.com/mitchellh/cli"
+	"reflect"
+	"strings"
+)
+
+// TODO: Everything in this file needs better naming.
```
swalkinshaw, line 10: What exactly is the purpose of this file/namespaces? I know it hides plugin commands from the normal commands list but I don't really understand the "namespaces" part.

plugin/namespaces.go
TangRufus: This snippet no longer exist after refactoring.

cmd/passthrough.go
TangRufus: Good catch

plugin/finder.go
TangRufus: Good catch

cmd/passthrough.go
TangRufus: See: https://github.com/kubernetes/kubernetes/pull/76227#discussion_r277102122

I must admit I don't understand the implications of "replaces the old process with the new one"

swalkinshaw: Same. Well I know enough from looking into it before that there is a difference. Still not sure if it matters here but might as well stick with what they have 👍

plugin/namespaces.go
TangRufus: Consider a plugin command like this: `trellis-one-two-three`:

When we add the `one two three` command to `mitchellh/cli`, it knows `one` and `one two` namespaces are missing and generates `MockCommand` for them. See: https://github.com/mitchellh/cli/blob/5454ffe87bc5c6d8b6b21c825617755e18a07828/cli.go#L371-L380

If we do not hide those `MockCommand`, they will show up on the command list with empty synopsis and mix with the core command list:

```sh-session
$ ./trellis-cli
Usage: trellis [--version] [--help] <command> [<args>]

Available commands are:
// ...
    new           Creates a new Trellis project
    one
    provision     Provisions the specified environment
// ...

$ ./trellis-cli one
This command is accessed by using one of the subcommands below.

Subcommands:
    two


$ ./trellis-cli one two
This command is accessed by using one of the subcommands below.

Subcommands:
    three    Third party plugin: Forward command to trellis-one-two-three
```

---

Same reason why we need to hide both `venv` and `venv hook` when only `venv hook` was added in https://github.com/roots/trellis-cli/pull/114/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R125-R131


---

Current implementation [`6f5873d`](https://github.com/roots/trellis-cli/pull/144/commits/6f5873db72b87c5b471b7f24f1ae0e5ade404c28) hides both `one` and `two`. In some cases, the help text is useless:

```sh-session
$ ./trellis-cli one
This command is accessed by using one of the subcommands below.
```

TODO: Change it to hide top level namesapce (e.g: `one`) only.

swalkinshaw: This should be consistent with the normal commands right? So for example, if I unhide `venv` and `venv hook`, we get this (other commands omitted):

```plain
Usage: trellis [--version] [--help] <command> [<args>]

Available commands are:
    vault         Commands for Ansible Vault
    venv

Available plugins commands are:
    one two three
```

Yet the plugin shows `one two three`. The equivalent would be only showing `one` right? Basically to be consistent we should only show top-level commands?

One other inconsistency is just running the top-level command doesn't show the subcommand:

```plain
❯ PATH="$(pwd)/testing:$PATH" ./trellis-cli another
This command is accessed by using one of the subcommands below.
```

But accessing the second sub command does work:

```plain
❯ PATH="$(pwd)/testing:$PATH" ./trellis-cli one two
This command is accessed by using one of the subcommands below.

Subcommands:
    three    Third party plugin: Forward command to trellis-another-two-three
```

TangRufus: Update in [`747e8d2` (#144)](https://github.com/roots/trellis-cli/pull/144/commits/747e8d20eddd9cd945c1349539e606166d0fd34a) to show top-level commands in `$ trellis --help`?

```sh-session
$ trellis --help
Usage: trellis [--version] [--help] <command> [<args>]

Available commands are:
...
    valet         Commands for Laravel Valet
    vault         Commands for Ansible Vault


Available third party plugins commands are:
    one
    plugintest
    theplugin


$ trellis one --help
This command is accessed by using one of the subcommands below.

Subcommands:
    two


$ trellis one two --help
This command is accessed by using one of the subcommands below.

Subcommands:
    three    Third party plugin: Forward command to trellis-one-two-three
```


For `$ trellis one two three --help`, it invokes the underlying plugin command, i.e: `$ trellis-one-two-three --help`

TangRufus: Definition of `namespace` changed! 

Now: A namespace is the top level command.

Take these commands as examples:

```
$ trellis check
$ trellis valet link
$ trellis one two three
```

The namespaces are `check`, `valet` and `one`.

Help wanted: We need better namings.

TangRufus:
> I think I'd prefer to skip this for now. I don't want people getting confused which subcommand is actually being run and not knowing if it's some custom override or not.

[`5947a5e` (#144)](https://github.com/roots/trellis-cli/pull/144/commits/5947a5ea700ed5dd6af10325427405e60ecc5435)  prevents overriding core commands or adding any subcommands under core commands. 
For example, these plugin commands are invalid:

```
# Because core provides the same command
$ trellis check

# Because core provides `venv hook`, anything under `venv` is reserved for core only
$ trellis venv
$ trellis venv foo
$ trellis venv foo bar
```

plugin/register.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	}
+
+	// Separate plugin commands from core command lists
+	c.HiddenCommands = append(c.HiddenCommands, pluginNamespaces...)
+	// Append plugin command list to help text.
+	c.HelpFunc = helpFunc(c.Name, pluginNamespaces)
+}
+
+func namespacesFor(v reflect.Value) []string {
```
swalkinshaw, line 31: You can use a named return value parameter here if you want to avoid the `var` later. Personally I think it's a little clearer that the var being declared is the return value.

```go
func namespacesFor(v reflect.Value) (namespaces []string) {
}
```

note: this won't work with maps since their "zero" value is `nil` and not an empty map

plugin/register.go
```diff
@@ -0,0 +1,57 @@
...
+	m := v.MapKeys()
+
+	var namespaces []string
+	for _, v := range m {
+		name := fmt.Sprint(v)
+		namespace := strings.Split(name, " ")[0]
```
swalkinshaw, line 37:
```go
		namespace := strings.Split(v.String(), " ")[0]

```

`fmt.Sprint` is just calling `name.String()` under the hood so think it's better to call it directly.

plugin/register.go
```diff
@@ -0,0 +1,57 @@
...
+		namespaces = append(namespaces, namespace)
+	}
+
+	return unique(namespaces)
+}
+
+// unique de-duplicates a given slice of strings without
+// sorting or otherwise altering its order in any way.
+func unique(values []string) []string {
```
swalkinshaw, line 46:
```go
func unique(values []string) (newValues []string) {
```

swalkinshaw (commented):
Looks good overall! For naming, I think "namespaces" should just be "root commands"? 🤔

plugin/register.go
TangRufus: done in [`c241027` (#144)](https://github.com/roots/trellis-cli/pull/144/commits/c241027becb2641840f9245a88baa589cc235925)

plugin/register.go
TangRufus: done in [`c241027` (#144)](https://github.com/roots/trellis-cli/pull/144/commits/c241027becb2641840f9245a88baa589cc235925)

plugin/register.go
TangRufus: done in [`44045b0` (#144)](https://github.com/roots/trellis-cli/pull/144/commits/44045b0d0fdbe969b57270296c824a39264b2d26)

swalkinshaw:
Anything left on your own? Tests look great 🎉 

I'll give it a final review if you think it's ready

TangRufus:
Ready for review:

1. Need testing on Windows and/or WSL
1. Need testing on real linux desktops


Note: I didn't add [`c.Trellis.LoadProject()`](https://github.com/roots/trellis-cli/blob/5de0b2ba268fe497a66645f48e1c1be788c1d288/cmd/exec.go#L20) in `PassthroughCommand` because `shell-init` should have taken care of `virtualenv` activation when we `cd` into the folder.

For the future:

1. Investigate whether we should combine/unify `ExecCommand.Run` and `PassthroughCommand.Run`. They are basically doing the same thing
1. Handle overshadowed plugins, i.e: plugins with the same binary name

plugin/register_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"os"
+	"path/filepath"
+	"reflect"
+	"sort"
+	"strings"
+	"testing"
+	"trellis-cli/cmd"
+)
+
+const spyCommand = `#!/usr/bin/env bash
```
swalkinshaw, line 16: Neat idea!

swalkinshaw (approved):
Amazing work on this 🎉

PR #146, opened 2020-12-13
Rename module from `trellis-cli` to `github.com/roots/trellis-cli`; `go get -u -t ./...`

TangRufus:
Goal: Developers using `trellis-cli` as a library (e.g: plugin commands) doesn't need to [`replace`](https://golang.org/ref/mod#go-mod-file-replace)

Example: https://github.com/ItinerisLtd/trellis-cyberduck/blob/3e174c9bdf2297ccdb22b9f4be906e53e7ae2459/go.mod#L10 is no longer required.


swalkinshaw:
🎉 makes sense. This entire codebase would probably use a re-org tbh since that's one of the parts of Go I know the least about.

swalkinshaw (approved):

PR #150, opened 2020-12-22
Update provision command help text

MikeiLL:
It took me a little while to find the syntax for multiple arguments. And more than once... I think this might make it a little clearer.

### DIFF
--- a/cmd/provision.go
+++ b/cmd/provision.go
@@ -98,11 +98,6 @@ See https://roots.io/trellis/docs/remote-server-setup/#provision for more detail
 Provision the production environment:
 
   $ trellis provision production
-  
-Options:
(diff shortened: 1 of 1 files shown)


cmd/provision.go
```diff
@@ -98,6 +98,15 @@ See https://roots.io/trellis/docs/remote-server-setup/#provision for more detail
 Provision the production environment:
 
   $ trellis provision production
+  
+Options:
```
swalkinshaw, line 102: I'd prefer to keep the options at the bottom for now just because it's how every command does it so its consistent. Want to just make the change below?

MikeiLL: Makes sense. I have done so. I forget if I need to resubmit the pull request or what. I think it updates automatically.

swalkinshaw:
Thanks!

PR #157, opened 2021-03-09
Use go 1.16 embed

swalkinshaw:
Uses Go's new 1.16 embed package to extract our static content into actual files instead of just strings.

Ref: https://golang.org/pkg/embed/


swalkinshaw:
@TangRufus I went to update the docker image but I don't have permission 😅

TangRufus:
Thanks! 🎉

PR #162, opened 2021-03-27
Add option for droplet image (default Ubuntu 20.04)

### LINKED ISSUE #161: Allow specifying digital ocean image
* [x] This request isn't a duplicate of an existing request

## Summary

Allow specifying which image to use when creating digital ocean droplet, as with `doctl` command, such as `--image ubuntu-20-04-x64`

## Motivation

Local development image (vagrant) is now Ubuntu 20-04-x64; I would like my remote server to be the same

## Additional context

Perhaps the value, by default, could be taken from/somehow linked to a value in trellis
- eg. where this is already being used to specify the vagrant box to spin up in development, like `vagrant.default.yml`
- or held elsewhere in a config file such as `wordpress_sites.yml`

swalkinshaw:
Fixes #161

Adds the `--image` option flag to the `droplet create` command and bumps the default to `ubuntu-20-04-x64`

### DIFF
--- a/cmd/droplet_create.go
+++ b/cmd/droplet_create.go
@@ -19,9 +19,9 @@ import (
 	"github.com/mitchellh/cli"
 	"github.com/mitchellh/go-homedir"
 	"github.com/posener/complete"
-	"golang.org/x/crypto/ssh"
 	"github.com/roots/trellis-cli/digitalocean"
 	"github.com/roots/trellis-cli/trellis"
+	"golang.org/x/crypto/ssh"
 )
 
 var client *digitalocean.Client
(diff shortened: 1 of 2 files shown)


TangRufus:
Should https://github.com/roots/trellis-cli/blob/83f6de5c1303ec8c4ff1f527e275c1223b6ba148/digitalocean/main.go#L33 be using the new flag?

swalkinshaw:
Weird yes... I had that change 🤔  no idea but happened but good catch since it's useless without it!

TangRufus (approved):

PR #171, opened 2021-05-24
add windows installation instructions

kas5986:
windows instructions were missing for manual install


swalkinshaw:
I don't use Windows, but makes sense to me 👍  thanks

PR #209, opened 2021-11-22
Fix provision in dev for Vagrant

swalkinshaw:
The `provision` command used to always run the `server.yml` playbook regardless of environment which isn't correct for development where `dev.yml` should be used.

This sets `dev.yml` as the playbook for development and also specifies Vagrant's generated inventory file if it exists. This is needed for Vagrant boxes to use the proper user, host, port and private SSH key.

### DIFF
--- a/cmd/provision.go
+++ b/cmd/provision.go
@@ -2,12 +2,16 @@ package cmd
 
 import (
 	"flag"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/mitchellh/cli"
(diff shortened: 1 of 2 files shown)


swalkinshaw:
I kind of thought about that briefly... maybe it was just laziness, but running this command already builds all the necessary `ansible-playbook` arguments which we'd have to change for `vagrant provision`. So this way, both environments can share the same command (other than the playbook file).

I don't feel strongly either way tbh

PR #210, opened 2021-11-26
Replace Sequel Pro with Sequel Ace

huubl:
See motivation:
https://github.com/roots/docs/pull/318


huubl:
Maybe adding?

```
case "sequel-pro":
    return &DBOpenerSequelAce{ui: ui}, nil
}
```
or
```
case "sequel-pro":
    return nil, fmt.Errorf("Sequel Pro is replaced by Sequel Ace. Check the docs for more info.")
}
```

https://github.com/roots/trellis-cli/blob/4d818540efa560bef04eaa16f7b78b77d22a21b2/cmd/db_opener_factory.go#L14-L23

swalkinshaw:
I like the second option 👍

swalkinshaw (approved):

swalkinshaw:
Thanks @huubl

PR #220, opened 2021-12-12
Improve init command

swalkinshaw:
Mainly this avoids showing the newly added warning about the virtualenv not existing while running the `init` command. If the warning tells users to run `trellis init`, then we shouldn't show the warning _again_ once they run it.

This also improves the output a bit by removing a redundant pip command line. There's also a final error message as well if `pip` fails.

Before (success):
image

Before (error):
image

After (success):
image

After (error):
image

### DIFF
--- a/main.go
+++ b/main.go
@@ -42,6 +42,7 @@ func main() {
 
 	ui := &cli.ColoredUi{
 		ErrorColor: cli.UiColorRed,
+		WarnColor:  cli.UiColor{Code: int(color.FgYellow), Bold: false},
 		Ui: &cli.BasicUi{
 			Reader:      os.Stdin,
 			Writer:      os.Stdout,
--- a/trellis/trellis.go
+++ b/trellis/trellis.go
@@ -3,7 +3,7 @@ package trellis
 import (
 	"errors"
 	"fmt"
-	"github.com/fatih/color"
+	"github.com/mitchellh/cli"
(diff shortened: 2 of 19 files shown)


main.go
```diff
@@ -42,6 +42,7 @@ func main() {
 
 	ui := &cli.ColoredUi{
 		ErrorColor: cli.UiColorRed,
+		WarnColor:  cli.UiColor{Code: int(color.FgYellow), Bold: false},
```
swalkinshaw, line 45: cli's yellow uses `color.FgHiYellow` which doesn't work in my terminal at least 🤔 got it around by doing this manually

trellis/trellis.go
```diff
@@ -103,21 +103,10 @@ func (t *Trellis) LoadProject() error {
 
 	os.Chdir(t.Path)
 
-	if os.Getenv("TRELLIS_VENV") != "false" {
+	if os.Getenv(TrellisVenvEnvName) != "false" {
 		if t.Virtualenv.Initialized() {
+			t.VenvInitialized = true
 			t.Virtualenv.Activate()
-		} else {
-			color.Yellow(`
```
swalkinshaw, line 110: Refactored this because using anything other than `cli.Ui` for output will cause this output in _all_ tests which is super annoying and not good.

Now, each command just needs to do the check manually which might be better in the end anyway.

PR #228, opened 2021-12-21
Add open command

### LINKED ISSUE #76: `open` in browser command
* [x] This request isn't a duplicate of an existing request

## Summary

I propose adding a new command, `trellis open`, to allow users to open the website in a browser. The full command schema would probably be something like:

`trellis open [options] ENVIRONMENT SITE`

## Motivation

`trellis open` would open the default environment in a browser. Inspiration for this is Laravel Valet and Netlify’s `open` command which do exactly that (the former for development, the latter for production).

## Additional context

Some things to consider would be:

* Which environment should be considered default? I think `development` should be, since I could see the command being most used in this environment.
* How does this behave for when multiple sites are on the same server? Of course, multiple sites on the same server isn’t ideal, but it is possible with Trellis. It could offer an selection menu, so that the command wouldn’t be useless in this context.
* An idea to tag on to this would be something like Netlify’s `open:admin` command which opens the admin dashboard for the website. It could offer to install [aaemnnosttv/wp-cli-login-command](https://github.com/aaemnnosttv/wp-cli-login-command) as well to make the transition more seamless. We could always just name this command `login`.

swalkinshaw:
Fixes #76 

`trellis open` is a way to easily open user-defined project specific "shortcuts" (usually URLs).

By default it opens the main site's canonical development URL in your default web browser. Any additional shortcuts can be defined in the CLI config in `.trellis/cli.yml` as a simple map of name to value.

For example:

```yaml
open:
  sentry: https://myapp.sentry.io
```

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -1,0 +1,149 @@
+package cmd
+
+import (
+	"fmt"
+	"runtime"
+	"sort"
+	"strings"
+
+	"github.com/mitchellh/cli"
+	"github.com/posener/complete"
+	"github.com/roots/trellis-cli/trellis"
+)
+
+type OpenCommand struct {
+	UI      cli.Ui
+	Trellis *trellis.Trellis
+}
+
+func (c *OpenCommand) Run(args []string) int {
+	if err := c.Trellis.LoadProject(); err != nil {
+		c.UI.Error(err.Error())
+		return 1
+	}
+
+	commandArgumentValidator := &CommandArgumentValidator{required: 0, optional: 1}
+	commandArgumentErr := commandArgumentValidator.validate(args)
+	if commandArgumentErr != nil {
+		c.UI.Error(commandArgumentErr.Error())
+		c.UI.Output(c.Help())
+		return 1
+	}
+
+	var openArgs = []string{}
+
+	if len(args) == 0 {
+		siteName, siteNameErr := c.Trellis.FindSiteNameFromEnvironment("development", "")
+		if siteNameErr != nil {
+			c.UI.Error(siteNameErr.Error())
+			return 1
+		}
+
+		url := c.Trellis.SiteFromEnvironmentAndName("development", siteName).MainUrl()
+		openArgs = []string{url}
+	} else {
+		value, exists := c.Trellis.CliConfig.Open[args[0]]
+
+		if !exists {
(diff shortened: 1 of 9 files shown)


cmd/open.go
```diff
@@ -0,0 +1,130 @@
...
+		if !exists {
+			c.UI.Error(fmt.Sprintf("Error: shortcut '%s' does not exist. Check your .trellis/cli.yml config file.", args[0]))
+			c.UI.Error(fmt.Sprintf("Valid shortcuts are: %s", strings.Join(openNames(c.Trellis.CliConfig.Open), ", ")))
+			return 1
+		}
+
+		openArgs = []string{value}
+	}
+
+	open := execCommand("open", openArgs, c.UI)
```
knowler, line 55: Does this work on Linux? Or do we need to use `xdg-open` for that?

swalkinshaw: I think I just naively assumed it exists.. but yeah looks like `xdg-open` is what we want. Added a conditional runtime check

knowler (approved):
This checks off my boxes for the original feature request. I haven’t actually tested this (nor will I be able to).

cmd/open.go
TangRufus: Perhaps https://github.com/cli/browser is a better choice?

Example: https://github.com/cli/browser/blob/38f8067bcf5aa9757803ff65d054299f244c0f87/examples/Open/main.go

TangRufus:
Question: Should we allow "variables" in `.trellis/cli.yml`?

For example:

```yaml
# .trellis/cli.yml

open:
  sentry: "https://{{ site.sentry.app_name }}.sentry.io"
```

```yaml
# group_vars/<env>/wordpress_sites.yml

wordpress_sites:
  example.com:
    sentry:
      app_name: my-app
```

cmd/open.go
swalkinshaw: Oh good find, yeah might be better to use that!

swalkinshaw:
re: variables... maybe? But I'll leave that for later and see if it not having them is really a problem. 

In your example I'm not sure it's really correct to have ` group_vars/<env>/wordpress_sites.yml` be the source of truth for _everything_. Especially in this case when it's CLI only information.

cmd/open.go
swalkinshaw: Can't really use that library unfortunately since it means we can't easily test this command as there's no control over the underlying `cmd`. Running a test literally opens a browser window locally 😂

PR #246, opened 2022-01-10
Improve alias command output

swalkinshaw:
Hides Ansible playbook output (since it's an implementation detail) and uses a spinner since it can take a few seconds.

image

@TangRufus should we just append this to `wp-cli.local.yml` automatically?

### DIFF
--- a/cmd/alias.go
+++ b/cmd/alias.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/fatih/color"
 	"github.com/mitchellh/cli"
 	"github.com/posener/complete"
 	"github.com/roots/trellis-cli/command"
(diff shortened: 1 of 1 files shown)


swalkinshaw:
We can look into appending to wp-cli.local.yml automatically separately

TangRufus:
> should we just **append** this to wp-cli.local.yml automatically?

Probably not. The idea of having a separate `wp-cli.trellis-alias.yml` is to ensure users do not mess with our generated file AND we don't mess with users' original wp-cli config files.

If we are to append this to wp-cli.local.yml, we need to:
1. wrap our alias config in some `# Trellis-CLI generated. DO NOT MODIFY!` comments
2. ensure we only change our alias config when updating, i.e: do not mess with users' original configs

### Option A

Search and replace between the `do not modify` comments. 

I can foresee users modifying our alias config without aware of it. VS code auto formatting the indentation might work but we are going to break something when we search and replace.

 ### Option B
 
1. Keep `do not modify` comments 
2. Read wp-cli.local.yml via a yaml parser
3. marshal the whole wp-cli.local.yml with new alias config (i.e: rewriting the whole wp-cli.local.yml)

Less error prone. OCD users will hate us changing all their styling (indentations, hyphen arrays VS square bracket arrays, etc).

swalkinshaw:
Makes sense 👍  let's keep it as is

PR #247, opened 2022-01-11
Hide output for internal playbooks

swalkinshaw:
Improves the output of `dotenv` and `db open` commands


swalkinshaw:
https://github.com/roots/trellis-cli/pull/250 replaces this

PR #250, opened 2022-01-16
Refactor exec command usage

swalkinshaw:
This refactors how `exec.Command` is used and called throughout the codebase. Before we had two functions:

* `execCommand` - hardcoded to output to `os.Stdout`
* `execCommandWithOutput` - hardcoded to output to `cli.Ui`

They were confusingly named and not flexible. Both used different io writer's to print their output, but *both* ended up with the same result and would output to the terminal. They also *both* resulted in "logging" the command being run.

Whenever we didn't want to show output in the terminal, we'd either have to use `cli.NewMockUi` or call `exec.Command` directly. The latter option meant that we couldn't properly test those commands either. I've never been happy with the existing code so this finally improves it.

This PR replaces those two methods with a new package `command` and a `Cmd` method to create the underlying `exec.Cmd`. There are descriptive options which allow any caller to apply some common options:

* `WithTermOutput` (uses `os.Stdout`)
* `WithUiOutput` (uses `cli.Ui`)
* `WithLogging` (outputs the command being run "Running command => ... ")

With options:
```go
cmd := command.WithOptions(command.WithTermOutput(), command.WithLogging(c.UI)).Cmd("ansible-vault", vaultArgs)
```

Without options:
```go
cmd := command.Cmd("ansible-vault", vaultArgs)
```

`Playbook` and `AdHocPlaybook` had to be changed as well to work with this refactor since using those meant that CLI commands wouldn't directly call `command.Exec`.


swalkinshaw:
I've updated this since I was able to figure out the interface I wanted. Also renamed some methods (like `Exec` to `Cmd`) to better reflect what they do.

With options:
```go
cmd := command.WithOptions(command.WithTermOutput(), command.WithLogging(c.UI)).Cmd("ansible-vault", vaultArgs)
```

Without options:
```go
cmd := command.Cmd("ansible-vault", vaultArgs)
```

PR #265, opened 2022-03-04
Readme: Bump go requirement to v1.15

TangRufus:
Because [`T.TempDir`](https://pkg.go.dev/testing#T.TempDir) used in #264 was added in go v1.15.

[ci skip]


TangRufus:
Replaced by #266

PR #269, opened 2022-03-06
Update Dockerfile: go 1.17, bullseye, python3

swalkinshaw:
This also switches to pip for Trellis dependency management to ensure the proper versions of Ansible are installed.

Note: I already updated the docker images using this Dockerfile, just forgot to commit this.

cc @TangRufus


TangRufus:
Looks good to me

PR #270, opened 2022-03-08
Enable --debug

E-VANCE:
Add `--debug`-flag to enable detailed logging of `vagrant up` ([Vagrant docs](https://www.vagrantup.com/docs/other/debugging)).

We could also use the `VAGRANT_LOG=info`-level as it is *quieter* but would require a slightly different approach since the commands differ and multiple steps are required on Windows.


swalkinshaw:
Thanks @E-VANCE

PR #288, opened 2022-04-29
Add GitHub secret and deploy key to specified `--repo`

### LINKED ISSUE #286: Feature Request: specify owner/repo for `key generate`
### Terms

- [X] I have read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
- [X] This request is not a duplicate of an existing issue
- [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

### Summary

Currently, when running `trellis key generate`, `gh secret set` will automatically pick the remote to set the secret to.
This is great for most cases, but allowing a user to specify a remote to set the secret on would be helpful to some.

#### Why are we doing this?
To allow setting a secret to a specific remote.

#### What use cases does it support?
If you have more than 1 remote (e.g. you have an `origin` remote and an `upstream` remote pointing to your maintained Trellis fork) then running `trellis key generate` (from my tests) always sets the secrets on the wrong remote.
This feature request would allow explicitly specifying the correct remote or owner/repo. 

#### What is the expected outcome?
Running `trellis key generate` with an optional flag to specify the remote or owner/repo would set the secrets to the expected repo.

#### Potential conflicts / foreseeable issues
None that I can think of; not specifying the remote repo would just result in the current behaviour.

### Additional Context

We use separate Trellis and Bedrock repositories for our preferred organisation of projects.

Also, we have a [tool we use internally](https://github.com/ItinerisLtd/iroots) to spin up projects from our maintained Bedrock and Trellis forks then configure and deploy with trellis-cli.

I plan to add support to the tool for automating `trellis key generate` during the project setup.
However, this is not currently possible due to the local repositories have 2 remotes and `trellis key generate`/`gh secret set` using on the wrong remote.

codepuncher:
Closes #286 

I have tested this with specifying: no repo (works as it did before), with an existing repo (works as intended by #286) and with non-existing repo.

For the non-existing repo, `gh api` returns `failed to fetch public key: HTTP 404: Not Found (https://api.github.com/repos/xxx/xxx/actions/secrets/public-key)` and is printed to console by existing `fmt.Errorf`.

### DIFF
--- a/cmd/key_generate.go
+++ b/cmd/key_generate.go
@@ -41,6 +41,7 @@ type KeyGenerateCommand struct {
 	noGithub     bool
 	path         string
 	provisionEnv string
+	repo         string
 }
 
 func (c *KeyGenerateCommand) init() {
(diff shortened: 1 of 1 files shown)


swalkinshaw:
🎉  looks good! Thank you

PR #292, opened 2022-05-02
Skip provisioning after generating keys

### LINKED ISSUE #291: Feature Request: Skip provisioning after generating keys
### Terms

- [X] I have read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
- [X] This request is not a duplicate of an existing issue
- [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

### Summary

Skip environment provisioning with a flag e.g. `--no-provision`.
Useful when using Trellis only for deployment (e.g. on managed hosts like Kinsta).

#### Why are we doing this?
To allow skipping environment provisioning after a key has been generated.

#### What use cases does it support?
For those using Trellis with managed hosts, there is no need for provisioning since the host is not configured by Trellis.
Or perhaps there is another step in your workflow that needs to happen before provisioning, so skipping this would be beneficial.

#### What is the expected outcome?
To skip provisioning the host.

#### Potential conflicts / foreseeable issues
N/A

### Additional Context

_No response_

codepuncher:
Closes #291

### DIFF
--- a/cmd/key_generate.go
+++ b/cmd/key_generate.go
@@ -39,6 +39,7 @@ type KeyGenerateCommand struct {
 	keyName      string
 	knownHosts   string
 	noGithub     bool
+	noProvision  bool
 	path         string
 	provisionEnv string
 	repo         string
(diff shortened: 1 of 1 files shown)


cmd/key_generate.go
```diff
@@ -276,14 +283,19 @@ Specify a repository to add the GitHub secret and deploy key to:
...
-      --path        Path for private key (Default: $HOME/.ssh)
-      --provision   Name of environment to provision after key is generated
-      --repo        Repository to add the GitHub secret and deploy key to (Format: OWNER/REPO)
-  -h, --help        show this help
+      --known-hosts  Comma-separated list of SSH known hosts (optional)
+      --name         Name of SSH key (Default: trellis_<site_name>_ed25519)
+      --no-github    Skips creating a GitHub secret and deploy key
+      --no-provision Skips provisioning the environment after key is generated
+      --path         Path for private key (Default: $HOME/.ssh)
+      --provision    Name of environment to provision after key is generated
```
codepuncher, line 296:
```suggestion
      --path         Path for private key (Default: $HOME/.ssh)
      --[no-]provision    Name of environment to provision after key is generated
```
Alternative help output. Thoughts?

swalkinshaw: I'd be okay with that 👍

swalkinshaw:
Thanks 👏

PR #311, opened 2022-08-02
Add certificate install and uninstall commands

swalkinshaw:
`trellis certificate install` will install the root certificate from a Trellis development VM (by default) into your computer's system truststore. This means local development HTTPS sites will be considered secure by web browsers and won't show insecure warnings.

`trellis certificate uninstall` removes the previously installed certificate from your computer's system "truststore".

By default this integrates with Trellis and relies on the changes in https://github.com/roots/trellis/pull/1310. However, by passing the `--path` option this is a useful standalone tool to trust *any* root certificate on your computer.


TangRufus (approved):
Looks good to me

hirasso:
Hi all! I'm late to the party as it seems. Is there an official status on trellis+lima? I'm seeing a few PRs like this one that have been open for ~2 years. No pressure, just asking :)

swalkinshaw:
https://github.com/roots/trellis/pull/1310 likely won't ever be merged at this point because it's a large refactor that risks breaking things and I'm not sure it's worth it for the benefits.

However, Trellis + Lima should be the default for macOS users at least. I just realized the `vm` commands are marked as experimental still so I'll remove that.

hirasso:
That's great to hear! Maybe someone should also give the article on the blog some love, so that newcomers realize this is now the recommended way of setting things up on MacOS. Thank you @swalkinshaw for your quick reaction!

PR #316, opened 2022-09-05
Add local env support to alias cmd

swalkinshaw:
Defaults to including the local env in `trellis alias` to include the Vagrant development environment.

For proper SSH support, this defaults to using the _main host_ which is what Trellis configures Vagrant to use as an SSH alias name (in your SSH config).

For non-Vagrant users, or any other non-default setup, `--skip-local` can be used to maintain the old behaviour.


TangRufus (approved):
Awesome!

PR #343, opened 2022-12-23
Add VM management via Lima

swalkinshaw:
Up until now, trellis-cli had limited integration with Vagrant for managing virtual machines. The `up` and `down` commands exist, but didn't fully wrap the underlying Vagrant usage.

Most of the Vagrant integration is done in Trellis itself, with the CLI just running `vagrant up` for you basically. It was quite limited, and still relied on Vagrant, which is quite slow and heavy for what we need.

This adds full integration for managing virtual machines in development via https://github.com/lima-vm/lima as an alternative to Vagrant. Lima stands for "Linux virtual machines (on macOS, in most cases)" and that's exactly our use case.

Lima isn't just a replacement for Vagrant, it also replaces the VM provider like VirtualBox or Parallels too. Lima supports two ways of running guest machines:

* [QEMU](https://www.qemu.org/)
* macOS Virtualization.Framework ("vz")

For this initial integration, we're only supporting the macOS Virtualization.Framework because it offers near-native performance, and that includes file syncing via `virtiofs`. 

For macOS Ventura (13.0+) users, we'll eventually recommend using trellis-cli's VM feature over Vagrant as the default since it's easier and faster. We'll look into expanding support for Linux users as well depending on performance of the file mounts on the VM.

## Requirements
* Intel or Apple Silicon
* macOS 13 (Ventura)
* Lima >= 0.14

## Usage
There's 5 new commands:

* `trellis vm start`
* `trellis vm stop`
* `trellis vm delete`
* `trellis vm shell`
* `trellis vm sudoers`

Under the hood, those commands wrap equivalent `limactl` features. Just like the previous Vagrant integration, you can always run `limactl` directly to manage your VMs.

For default use cases, `trellis vm start` can be run without any customization first. It will create a new virtual machine (using Lima) from a generated config file (`project/trellis/.trellis/lima/config/<name>.yml`). The site's `local_path` will be automatically mounted on the VM and your `/etc/hosts` file will be updated.

Note: see `trellis vm sudoers -h` for more details on making the `/etc/hosts` updates passwordless. Or just run:
```bash
$ trellis vm sudoers | sudo tee /etc/sudoers.d/trellis
```

## Configuration
Right now configuration options for the VM are limited. The intention is to the most common use cases without any configuration needed.

A trellis-cli config file (global or project level) supports a new `vm` option. The only useful config option right now is `images` (to switch between Ubuntu 20.04 and 22.04 for example). 20.04.

Here's an example of specifying Jammy 22.04 for ARM (`aarch64`) only, since that's what I use):

```yml
# .trellis/cli.yml
vm:
  images:
    - location: https://cloud-images.ubuntu.com/jammy/current/jammy-server-cloudimg-arm64.img
      arch: aarch64
```

### DIFF
--- a/app_paths/app_paths.go
+++ b/app_paths/app_paths.go
@@ -15,7 +15,11 @@ const (
 	xdgDataHome      = "XDG_DATA_HOME"
 )
 
-// Config path precedence: TRELLIS_CONFIG_DIR, XDG_CONFIG_HOME, AppData (windows only), HOME.
+// Config path precedence:
+// 1. TRELLIS_CONFIG_DIR
+// 2. XDG_CONFIG_HOME
+// 3. AppData (windows only)
+// 4. HOME
 func ConfigDir() string {
 	var path string
 
@@ -37,7 +41,10 @@ func ConfigPath(path string) string {
 	return filepath.Join(ConfigDir(), path)
 }
 
-// Cache path precedence: XDG_CACHE_HOME, LocalAppData (windows only), HOME.
+// Cache path precedence:
+// 1. XDG_CACHE_HOME
+// 2. LocalAppData (windows only)
+// 3. HOME
 func CacheDir() string {
 	var path string
 	if a := os.Getenv(xdgCacheHome); a != "" {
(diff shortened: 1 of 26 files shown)


retlehs:
Vagrant specific text that needs updating: https://github.com/roots/trellis/blob/dbc203747894f917d55c2a64d54e2a4d5d2ca468/lib/trellis/vagrant.rb#L73-L82

swalkinshaw:
@retlehs added that post-start message

swalkinshaw:
Closing in favour of https://github.com/roots/trellis-cli/pull/346

PR #346, opened 2023-01-07
Add VM integration

swalkinshaw:
Adds full integration for managing virtual machines in development via https://github.com/lima-vm/lima as an alternative to Vagrant. Lima stands for "Linux virtual machines (on macOS, in most cases)" and that's exactly our use case.

Lima isn't just a replacement for Vagrant, it also replaces the VM provider like VirtualBox or Parallels too. Lima supports two ways of running guest machines:

* [QEMU](https://www.qemu.org/)
* macOS Virtualization.Framework ("vz")

For this initial integration, we're only supporting the macOS Virtualization.Framework because it offers near-native performance, and that includes file syncing via `virtiofs`.

For macOS Ventura (13.0+) users, we'll eventually recommend using trellis-cli's VM feature over Vagrant as the default since it's easier and faster. We'll look into expanding support for Linux users as well depending on performance of the file mounts on the VM.

Requirements:
* Intel or Apple Silicon
* macOS 13 (Ventura)
* Lima >= 0.14

Usage:
There's 5 new commands:

* `trellis vm start`
* `trellis vm stop`
* `trellis vm delete`
* `trellis vm shell`
* `trellis vm sudoers`

Under the hood, those commands wrap equivalent `limactl` features. Just like the previous Vagrant integration, you can always run `limactl` directly to manage your VMs.

For default use cases, `trellis vm start` can be run without any customization first. It will create a new virtual machine (using Lima) from a generated config file (`project/trellis/.trellis/lima/config/<name>.yml`). The site's `local_path` will be automatically mounted on the VM and your `/etc/hosts` file will be updated.

Note: run `trellis vm sudoers -h` to make `/etc/hosts` file passwordless:
```bash
$ trellis sudoers | sudo tee /etc/sudoers.d/trellis
```

Configuration:
Right now configuration options for the VM are limited. The intention is to the most common use cases without any configuration needed.

A trellis-cli config file (global or project level) supports a new `vm` option. The only useful config option right now is `ubuntu` (to switch between Ubuntu 20.04 and 22.04 for example).

Here's an example of specifying Jammy 22.04:

```yml
vm:
  ubuntu: 22.04
```

### DIFF
--- a/app_paths/app_paths.go
+++ b/app_paths/app_paths.go
@@ -15,7 +15,11 @@ const (
 	xdgDataHome      = "XDG_DATA_HOME"
 )
 
-// Config path precedence: TRELLIS_CONFIG_DIR, XDG_CONFIG_HOME, AppData (windows only), HOME.
+// Config path precedence:
+// 1. TRELLIS_CONFIG_DIR
+// 2. XDG_CONFIG_HOME
+// 3. AppData (windows only)
+// 4. HOME
 func ConfigDir() string {
 	var path string
 
@@ -37,7 +41,10 @@ func ConfigPath(path string) string {
 	return filepath.Join(ConfigDir(), path)
 }
 
-// Cache path precedence: XDG_CACHE_HOME, LocalAppData (windows only), HOME.
+// Cache path precedence:
+// 1. XDG_CACHE_HOME
+// 2. LocalAppData (windows only)
+// 3. HOME
 func CacheDir() string {
 	var path string
 	if a := os.Getenv(xdgCacheHome); a != "" {
@@ -47,6 +54,23 @@ func CacheDir() string {
 	} else {
 		c, _ := os.UserHomeDir()
 		path = filepath.Join(c, ".local", "state", "trellis")
+	}
+	return path
+}
+
+// Data path precedence:
+// 1. XDG_DATA_HOME
+// 2. LocalAppData (windows only)
+// 3. HOME
+func DataDir() string {
+	var path string
+	if a := os.Getenv(xdgDataHome); a != "" {
+		path = filepath.Join(a, "trellis")
+	} else if b := os.Getenv(localAppData); runtime.GOOS == "windows" && b != "" {
+		path = filepath.Join(b, "Trellis CLI")
+	} else {
+		c, _ := os.UserHomeDir()
+		path = filepath.Join(c, ".local", "share", "trellis")
 	}
 	return path
 }
--- a/cli_config/cli_config.go
+++ b/cli_config/cli_config.go
@@ -11,6 +11,18 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+type VmImage struct {
+	Location string `yaml:"location"`
+	Arch     string `yaml:"arch"`
+}
+
+type VmConfig struct {
+	Manager       string    `yaml:"manager"`
+	HostsResolver string    `yaml:"hosts_resolver"`
+	Images        []VmImage `yaml:"images"`
+	Ubuntu        string    `yaml:"ubuntu"`
+}
+
 type Config struct {
 	AllowDevelopmentDeploys bool              `yaml:"allow_development_deploys"`
 	AskVaultPass            bool              `yaml:"ask_vault_pass"`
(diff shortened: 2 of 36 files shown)


swalkinshaw:
https://github.com/lima-vm/lima/releases/tag/v0.15.0 is finally out so I'm going to merge this now.

Note: this is still under "experimental" status until we get more people testing it out.

artshostak:
Screenshot 2023-03-09 at 10 32 38 AM

Following the guide above now to try this out, but quick note the command above should be:
`trellis vm sudoers | sudo tee /etc/sudoers.d/trellis`

artshostak:
Just logging my experience here in case it helps someone else trying this out or aids the project with the documentation:

[1] I have `trellis-cli-dev` installed instead of `trellis-cli` — so my first move was to replace it:
`brew uninstall roots/tap/trellis-cli-dev`
`brew install roots/tap/trellis-cli`
This required a library "gcc" which can be installed with `brew install gcc` in full or apparently is quicker as part of XCode Command Line Tools, which can be installed with `xcode-select --install`

[2] Once I have trellis-cli version 1.10.0, I noticed my Ventura 13.1 didn't have Lima, so I had to install that as well:
`brew install lima`

[3] Created a new project with `trellis new {project}` and then gave it access to /etc/hosts with:
`trellis vm sudoers | sudo tee /etc/sudoers.d/trellis`

[4] The part that wasn't clear here is the next step to take, do I run `trellis up` the way I would to setup, or do I setup VM with `trellis vm start` but it looks like the latter proceeded to do first-time provisioning of the local development environment and created a `lima` folder in `trellis/.trellis` as anticipated. Was up and running in 7 minutes, definitely an improvement over the Parallels setup!

Note: Looks like step [3] was a little premature, and should have been done after step [4], so I did it again after with success.

Question here is if I wanted to specify my Ubuntu installation (20.04 instead of 22.04), when should I have done this prior to on a project level for trellis-cli config with the following:
```
vm:
  ubuntu: 20.04
```

The other question now is what happens if I have a project using Parallels and want to switch it over to VM midway for local development? Or do you not recommend that?

Thank you!

jkananen:
> * what happens if I have a project using Parallels and want to switch it over to VM midway for local development

I didn't get this right on the first go and had to start over a few times, but I think the necessary steps are:
- Latest trellis (1.20.1, or at least later than 1.18.0 that I first had)
- Removing / commenting out `vagrant-hostmanager-start` entries on your Mac's `/etc/hosts` file
- Updating the IPs in `trellis/hosts/development`. I'm not quite sure of this, but things started to work finally after I updated this one.
- `trellis vm start`
- Database migration / import in a way or another

> Just wondering how to tell trellis to use vm over vagrant?

Just use `trellis vm` commands.

swalkinshaw:
Thank you for the testing and feedback 😄 

I'll write more documentation on this feature soon including details on how to migrate or use it concurrently with Vagrant since you might run into `/etc/hosts` conflicts as you noticed.

One quick clarification: in most situations, you shouldn't need to manually edit the `hosts/development` file as it's not used. trellis-cli generates and manages an inventory/hosts file in `.trellis/lima` (just like Vagrant does in `.vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory`).

`trellis provision` has proper integration to know about those inventory files for example, but some other commands are still missing it which I'll work on improving. `trellis alias` for example will still use the wrong development host with Lima.

swalkinshaw:
Some replies to your questions @artshostak 

> ❌ Not able to access the shell in development with `trellis vm shell` produces the following error:
```
Running command => limactl shell {project}
exit status 255
```

If you run into that again, I'd recommend running the lima command directly. eg: `limactl shell --debug {project}` and see if there's more output to help troubleshoot.

> what happens if I have a project using Parallels and want to switch it over to VM midway for local development, or do you not recommend that? Just wondering how to tell trellis to use vm over vagrant?

As @jkananen said, right now `trellis vm` will only ever use the new VM feature with Lima. `trellis up` might be deprecated/removed at some point to reduce confusion (since `vagrant up` can easily be used; `trellis up` does nothing more than running that command).

> if I wanted to specify my Ubuntu installation (20.04 instead of 22.04), when should I have done this prior to on a project level for trellis-cli config with the following:
```
vm:
  ubuntu: 20.04
```

Before running `trellis vm start` - or by re-creating a new one after `trellis vm delete`.

> While it's blazing fast, it's unfortunately not so stable to do long sessions of local development with. Hope that helps and thanks again!

I wonder if all your stability/networking issues were related 🤔 I haven't heard of any issues like that yet. I'm wondering if a reboot helped (or would have)? If you run into that situation again, it would be good to confirm networking/DNS by continously `ping {dev host}` and see what comes back.

jkananen:
> One quick clarification: in most situations, you shouldn't need to manually edit the `hosts/development` file as it's not used.

Correct, it's not needed. It coincided earlier with rebooting my Mac, which seems to be necessary after the first `trellis vm start` run. This has been necessary now with all my three `trellis vm` setups, and the latest was otherwise very clear & straightforward.

Anyone else experiencing the need for a reboot?

swalkinshaw:
What makes you need to reboot?

I've had an issue before where it never got passed the SSH requirement and rebooting fixed it.

jkananen:
Browser fails to load the page and finally gives up with "This site can't be reached / ERR_CONNECTION_REFUSED"

swalkinshaw:
@jkananen next time that happens it would be useful to see:

1. you can shell into the VM
2. ping the hostname locally to see what IP is resolved
3. that `/etc/hosts` still looks correct

PR #371, opened 2023-03-21
Improve name options for vm start

swalkinshaw:
In some cases, trellis-cli could fail to determine the VM name and would error without any other options.

This provides two options:
1. a new `--name` flag to manually set it
2. a prompt for the name when the default fails


swalkinshaw:
Closing in favour of https://github.com/roots/trellis-cli/pull/373

PR #387, opened 2023-04-16
Support new trellis.cli.yml config file

swalkinshaw:
Adds more supported config files. In order of precedence, the following paths will be loaded:

* `.trellis/cli.yml` - existing file, now considered deprecated
* `trellis.cli.yml` - new path
* `trellis.cli.local.yml` - optional local override config

This change is being done to allow the `.trellis` directory to be gitigored as it contains CLI generated files that are machine specific and should not be committed to Git.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,15 +157,17 @@ Supported commands so far:
 There are three ways to set configuration settings for trellis-cli and they are
 loaded in this order of precedence:
 
-1. global config
-2. project config
-3. env variables
+1. global config (`$HOME/.config/trellis/cli.yml`)
+2. project config (`trellis.cli.yml`)
+3. project config local override (`trellis.cli.local.yml`)
+4. env variables
 
 The global CLI config (defaults to `$HOME/.config/trellis/cli.yml`)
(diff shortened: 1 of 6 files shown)


swalkinshaw:
I'm still considering other CLI config names. Here are some options:

* `cli.yml`
* `.cli.yml`

I think `trellis.yml` is too confusing right now since it doesn't mention CLI and all the current config settings are much more related to the CLI than Trellis itself.

And I'm 50/50 on whether to use the leading `.` to make it a hidden file or not. Since the file is optional (and additive), and not directly used by Trellis, the dot prefix makes that more obvious. But what's the actual point in hiding it? The leading `.` is more annoying to use and it makes it less visible.

swalkinshaw:
Got some [community feedback](https://discourse.roots.io/t/trellis-clis-new-config-file-poll/25178) and decided to go with `trellis.cli.yml`.

* it has CLI in the name to differentiate it from playbook files and make it obvious it's for the CLI
* not a hidden file which makes it more visible

PR #391, opened 2023-05-05
Update alias.go output message

MikeiLL:
In case other's like me don't know there's one built-in to Bedrock. I spent time wondering why the one I was putting above /site was being ignored.

### DIFF
--- a/cmd/alias.go
+++ b/cmd/alias.go
@@ -194,7 +194,7 @@ func (c *AliasCommand) Run(args []string) int {
 	spinner.Stop()
 	c.UI.Info("")
 	message := `
-Action Required: use the generated config by adding these lines to your wp-cli.yml or wp-cli.local.yml config.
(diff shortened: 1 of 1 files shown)


cmd/alias.go
```diff
@@ -194,7 +194,7 @@ func (c *AliasCommand) Run(args []string) int {
 	spinner.Stop()
 	c.UI.Info("")
 	message := `
-Action Required: use the generated config by adding these lines to your wp-cli.yml or wp-cli.local.yml config.
+Action Required: use the generated config by adding these lines to site/wp-cli.yml or an alternative wp-cli.yml/wp-cli.local.yml config.
```
swalkinshaw, line 197: I like the `site/` addition, but `wp-cli.yml/wp-cli.local.yml` could be confusing because the `/` there is for "slash" and not a path separator right?

Maybe say "wp-cli.yml (or wp-cli.local.yml)" ?

swalkinshaw:
Thank you!

PR #398, opened 2023-06-28
Fix use of reserved characters in wp-cli.yml

kellymears:
I don't _think_ the generated `wp-cli.yml` is compliant with the yml spec.

> The “@” (x40, at) and “`” (x60, grave accent) are reserved for future use.

> Reserved indicators can't start a plain scalar.

— [yaml spec (1.2.2)](https://yaml.org/spec/1.2.2/)

`yamllint` and vscode's default YAML checks report errors, so I think I might be on to something.

```txt
test.yml
  1:1       error    syntax error: found character '@' that cannot start any token (syntax)
```

### DIFF
--- a/cmd/files/playbooks/alias_template.yml.j2
+++ b/cmd/files/playbooks/alias_template.yml.j2
@@ -1,9 +1,9 @@
 {% if include_local_env | default(false) %}
-'@{{ env }}':
+"@{{ env }}":
   ssh: "{{ local_hostname_alias }}"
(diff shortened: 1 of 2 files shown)


swalkinshaw:
`alias_test.go` needs to be updated but otherwise this looks good 👍

chrillep:
@swalkinshaw when will this be released? 😅

swalkinshaw:
@chrillep I'll do one this week (maybe tonight).

Issue #402, opened 2023-07-12
`trellis vm shell` doesn't forward ssh agent


johnkraczek:
### Version

  Trellis v1.21.0   Bedrock v1.22.2

### What did you expect to happen?

After running: 

`trellis new example.com`
`cd example.com/trellis`
`trellis vm start`

Trellis provisions the lima VM

I run `trellis vm shell` which puts me into the VM. 
`echo "$SSH_AUTH_SOCK"` should return the host agent so that I can authenticate to ssh services. 

### What actually happens?

After doing the above, to work around this I can modify the ssh config and things work like they should. 
`limactl show-ssh -f config example.com > ~/.ssh/config`

that will put this into my config file:
```
Host lima-example.com
  IdentityFile "/Users/john/.lima/_config/user"
  IdentityFile "/Users/john/.ssh/id_rsa"
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null
  NoHostAuthenticationForLocalhost yes
  GSSAPIAuthentication no
  PreferredAuthentications publickey
  Compression no
  BatchMode yes
  IdentitiesOnly yes
  Ciphers "^aes128-gcm@openssh.com,aes256-gcm@openssh.com"
  User john
  ControlMaster auto
  ControlPath "/Users/john/.lima/example.com/ssh.sock"
  ControlPersist 5m
  Hostname 127.0.0.1
  Port 52058
```

I can edit the file and:
remove `ControlMaster`, `ControlPath`, and `ControlPersist` 
add `ForwardAgent yes`

Also I make sure that my key is available to the agent:
`ssh-add --apple-use-keychain ~/.ssh/[MY KEY]`

now if I `ssh lima-example.com`
It enters the shell and if I run `echo "$SSH_AUTH_SOCK"`
I am returned with `/tmp/ssh-XXXXRB6A0u/agent.24603`

Additionally if I attempt to ssh to bitbucket or github:

```
john@lima-example:~$ ssh johnkraczek@bitbucket.org
PTY allocation request failed on channel 0
authenticated via ssh key.

You can use git to connect to Bitbucket. Shell access is disabled
Connection to bitbucket.org closed.
```
As I would expect. 

I'm not sure if this is exactly an issue with trellis or an issue with lima-vm. I found this issue over there:
[SSH ForwardAgent does not work correctly](https://github.com/lima-vm/lima/issues/626)
[...]

PR #403, opened 2023-07-18
updated lima config ssh forwardAgent true by default

johnkraczek:
This solved the issue here
https://github.com/roots/trellis-cli/issues/402

Setting the ssh: forwardAgent: true in the lima config allows the agent to be forwarded. 
If you are good with that being the default then that's all that is needed.


swalkinshaw (approved):

swalkinshaw:
Thank you!

Issue #511, opened 2025-02-12
trellis logs for lima vm


lukasbesch:
### Summary

`trellis logs development` should adjust (omit) the ssh user automatically when using lima.

At the moment, the ssh user is hardcoded to `vagrant` when the environment is `development`:
https://github.com/roots/trellis-cli/blob/master/trellis/ssh.go#L10-L11

It should be possible to detect vagrant or lima automatically in most cases (by looking into the `.trellis` directory).
If both vagrant and lima configuration is present, it probably should assume vagrant (backwards compatible).
Possibly an additional command `trellis vm logs` would make sense for this edge case.

### Additional context

Output when running `trellis logs development`:

```
Running command => ssh vagrant@example.test tail -f /srv/www/example/logs/*[^gz]?
vagrant@example.test: Permission denied (publickey).
Error running ssh: exit status 255
```

Command needed to get the logs:
```
ssh example.test "tail -f /srv/www/example/logs/*[^gz]?"
```
Or enter the vm using `trellis vm shell` and tail from inside.

PR #525, opened 2025-03-12
Add compatibility for log-commands when using other DEV VM managers like Lima

E-VANCE:
This PR should add compatibility for running log-commands like `trellis logs --error development` when using other DEV VM managers then Vagrant like Lima.

See #511.

Also includes an added test case for ensuring the `tailCmd` method correctly generates command strings based on different flag combinations.

Couldn't get around to test this one yet as I am unsure how to use `brew` to install the forked **trellis-cli**...?

**Disclaimer:** I am not a `go`_-to-guy_ and had a little help from AI and there may be a few too many comments included 💡

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -80,6 +80,49 @@ func (c *LogsCommand) Run(args []string) int {
 		return 1
 	}
 
+	if environment == "development" {
+		// Use findDevInventory to auto-detect the VM provider
+		devInventory := findDevInventory(c.Trellis, c.UI)
+
+		// If we have a valid inventory and found a VM manager, use it
+		if devInventory != "" {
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Thanks for the proposed improvement.

> Couldn't get around to test this one yet as I am unsure how to use brew to install the forked trellis-cli...?

When you run `go build` locally, I just export the PATH like this:

```bash
export PATH=$PATH:/your/path/to/trellis-cli
```

Then run `trellis-cli` from anywhere to run the local binary

My main feedback for this PR is that I'd like to have the logic unified. Right now there's two completely separate paths. For example the `goaccess` check is duplicated in two places. Ideally we could figure out the tail command separately.

PR #598, opened 2025-10-22
Drop Vagrant support

retlehs:
Ref https://github.com/roots/trellis/pull/1622


cmd/check.go
```diff
@@ -24,23 +24,6 @@ var Requirements = []trellis.Requirement{
 			return strings.Replace(output, "Python ", "", 1)
 		},
 	},
-	{
```
swalkinshaw: Oh... guess we should add lima here? 🙈

help.go
```diff
@@ -39,7 +39,10 @@ func deprecatedCommandHelpFunc(commandNames []string, f cli.HelpFunc) cli.HelpFu
 		sort.Strings(keys)
 
 		for _, key := range keys {
-			commandFunc := commands[key]
+			commandFunc, ok := commands[key]
+			if !ok {
+				continue
+			}
```
swalkinshaw: Why did this change?

trellis/ssh.go
```diff
@@ -7,12 +7,8 @@ import (
 func (t *Trellis) SshHost(environment string, siteName string, user string) string {
 	host := t.SiteFromEnvironmentAndName(environment, siteName).MainHost()
 
-	if environment == "development" {
```
swalkinshaw: Ugh logs just never even worked on Lima. So this is fine anyway

help.go
retlehs: Needed for the `--help` command to work for some reason

cmd/alias.go
```diff
@@ -214,7 +214,7 @@ Usage: trellis alias [options]
 
 Generate WP CLI aliases for remote environments
 
-Don't include local env (useful for non-Vagrant users):
+Don't include local env (useful for non-Lima users):
 
   $ trellis alias --skip-local
```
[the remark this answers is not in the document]
swalkinshaw: Yeah but what's the point in even offering this option? Does it cause some issue you generated aliases for local/env without skipping it?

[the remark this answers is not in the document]
tangrufus: If I remember it correctly, there were problems when both `wp-cli.yml` and `wp-cli.trellis-alias.yml` specific the same alias. The config get merged / overwritten in ways we didn't expect.

PR #694, opened 2026-06-01
Rename SSH key option from --name to --key-name

stephan281094:
When running `trellis key generate --name "my_key"` it gave me the following error:
>flag provided but not defined: -name

Looking into this, I found that the command uses `--key-name` instead of `--name`. This PR ensures `--key-name` is used everywhere.

### DIFF
--- a/cmd/key_generate.go
+++ b/cmd/key_generate.go
@@ -264,7 +264,7 @@ To skip the GitHub specific parts, use the --no-github option:
 
 Specify a custom key name:
 
-  $ trellis key generate --name "my_key"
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Thanks for the fix and finding this issue. But thinking about it, I think just `name` is simpler? Here's the alternative for that: https://github.com/roots/trellis-cli/pull/695

swalkinshaw:
Went with https://github.com/roots/trellis-cli/pull/695. Thanks again

stephan281094:
Either option is fine for me. Thank you for the quick response and fix!

